PR #1, opened 2011-12-22
CCR <Body> node

snosrap:
This change makes the exported CCR validate against the schema. CCR does not contain a `<body>` node.


eedrummer:
Thanks for the fix!

PR #7, opened 2012-03-19
I believe the metadata objects, from and to_xml code is ready now

rdrand:
I was a little uncomfortable with one code snippet, but it should be ok for now. See the FIXME in record_date. It wasn't clear why Mongoid didn't allow self[:attr] or self.attr to work properly there for the embed_one Pedigree object.


rdrand:
OK, the necessary changes have been made and pushed to my branch

PR #9, opened 2012-04-24
Develop

jmazella:
Fixing the CCR exporter to handle a procedure with no time field.

### DIFF
--- a/lib/health-data-standards/export/ccr.rb
+++ b/lib/health-data-standards/export/ccr.rb
@@ -273,12 +273,14 @@ module HealthDataStandards
             patient.procedures.each_with_index do |procedure, index|
               xml.Procedure do
                 xml.CCRDataObjectID("PR000#{index + 1}")
-                xml.DateTime do
-                  xml.Type do
-                    xml.Text("Service date")
+                if !procedure.time.nil?
+                  xml.DateTime do
+                    xml.Type do
+                      xml.Text("Service date")
+                    end
+                    #time
+                    xml.ExactDateTime(convert_to_ccr_time_string(procedure.time))
                   end
-                  #time
-                  xml.ExactDateTime(convert_to_ccr_time_string(procedure.time))
                 end
                 xml.Description do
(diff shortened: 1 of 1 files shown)


projectcypress:
reply+i-4268474-4035f1bfa190225a0e7f00d8d4f8e6a9cb45fd91-1188796@REPLY.GITHUB.COM

rdingwell:
Did you make sure that this still runs the test suit WITH the CCR schema involved in the testing?

rdingwell:
This should also probably use the as_point_in_time method on the entry.  This looks for the time in the various places it may be for the entry.

rdingwell:
One additional thing.  This is a problem in more than just the procedures, it;s in conditions , results , .....  Can you please extrapolate this out into a method that will put the DateTime in the output.  Than update the template sections to just call that method with the current xml builder and the time you want to put in. 

 def to_ccr_date(xml, time)
    if time
         xml.DateTime do
                  xml.Type do
                    xml.Text("Initial Occurrence")
                  end
                  #time
                  xml.ExactDateTime(convert_to_ccr_time_string(time))
                end
          end
end

jmazella:
I changed it like you asked, looks like its working. The CCR export unit test wasn't actually testing anything even if the schema was present, so I replaced the #todo in there with an actual validation check.

rdingwell:
Have you been keeping your repo in sync with the cypress repo?  There is a test in there that does schema validation for CCR export.

jmazella:
Where is that?

rdingwell:
It's in the develop branch in the unit/export/ccr_test.rb

https://github.com/projectcypress/health-data-standards/blob/develop/test/unit/export/ccr_test.rb

jmazella:
oooooh, I'm smart. I was only thinking of the cypress tests. Yes, the health data standards tests pass with the new code.

rdingwell:
yes, but does it  pass with the schema validation enabled?

jmazella:
Yep

PR #11, opened 2012-06-14
Moving to utc time for everything

rdingwell:
Went through the code and moved to use utc time everywhere I could find that we were not


projectcypress:
reply+i-5070507-b69ee71355c93fa037a7c51406b650310942bd94-1188796@REPLY.GITHUB.COM

PR #13, opened 2012-06-20
Develop

jletourneau80:
update to include diagnosis active priority principal in export and import

### DIFF
--- a/lib/health-data-standards/export/view_helper.rb
+++ b/lib/health-data-standards/export/view_helper.rb
@@ -35,7 +35,8 @@ module HealthDataStandards
           '413322009'
         end
       end
-      
+
+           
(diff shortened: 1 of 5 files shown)


projectcypress:
reply+i-5182406-43a7480508a443c873f0ddea9fce4c96a6a76b68-1188796@REPLY.GITHUB.COM

abgoldstein:
I'm accepting this pull request with just a small modification to pull out puts statements.

The reason for the subsequent test breaks in Cypress did turn out to be a fixture issue. For whatever reason, the Record fixture that we were exporting had its Conditions listed as vanilla Entries so it didn't know about the specific ordinality method. Everything looks good on both ends now.

PR #15, opened 2012-06-27
Develop

rdingwell:
I updated the import export code to deal with value elements that are not just physical quantities.  Export now looks to see if something is a number before it creates a PQ type.  If it is not a number it checks if the value is either true or false and creates a BL type if it is, otherwise the catch all for leftovers is an ST type.


abgoldstein:
This looks good to me. I'm merging the request with minor modifications to pull out puts statements and flesh out a couple of comments.

PR #20, opened 2012-11-19
Add address and telcoms to patient record

hadleynet:
Added address and telecoms to record model. Added support for address and model to C32 and HTML exporters. Added tests.


hadleynet:
Also added C32 importer support for address and telecoms.

PR #22, opened 2012-12-19
GC32 Updates

ssayer:
A bunch of updates from an older RHEx branch, mostly involving GC32.


eedrummer:
This has been merged into condensing

PR #24, opened 2013-01-29
Procedure Order Import Changes

nonlinearity:
Changes needed for importing Procedure Order (2.16.840.1.113883.10.20.24.3.63)


hadleynet:
Do we capture that this is an order for a procedure rather than an actual procedure that happened anywhere ?

nonlinearity:
Good point.  I don't think it does explicitly.  I will make sure the 'ordered' value gets into the status_code field and submit another pull request.

nonlinearity:
All set.

2nd commit pending approval with the initial pull request:  procedure_order_import.rb now includes the status_code key 'HL7 ActStatus' with the value 'ordered'. This conforms to what we see in examples of patients with procedures ordered (Andre pulled a JSON example out of Bonnie for me to be sure).

PR #26, opened 2013-01-31
Importers added: Device Applied, Procedure Result, Risk Category Assessm...

nonlinearity:
...ent, Symptom Active


eedrummer:
Merged into develop

PR #29, opened 2013-02-14
QRDA Cat 1 Import Refactoring

nonlinearity:
EntryPackage object created - encapsulates all code for creating a complete QRDA entry with HQMF oid and proper status
Added assertions for status codes and hqmf oids to QRDA import unit tests
Rewrote external Cat1 importer class test cases to leverage EntryPackage object
Wrote test case for care goals (hand-coded test XML fragment using QRDA spec)
Fixed JSON HQMF lookup file error


nonlinearity:
Good points -- I'll fix the code once you've finished reviewing it!

nonlinearity:
Ok I will implement your suggested modifications.

PR #33, opened 2013-04-04
NLM API Updates

nonlinearity:


eedrummer:
Is there a reason you are setting the effective date in the constructor as opposed to on the get_valueset method?

nonlinearity:
I liked it better this way as far as code organization goes, but I can see the benefit of being able to call it with a variety of effective dates without having to reinstantiate the object.  I'll move it.

andrequina:
Can you send me a mongo dump of the latest VS loaded via the service?  I want to cross reference against what was loaded from the XLS

PR #47, opened 2013-08-14
use CodeSystemHelper to normalize code system names

carpeliam:


marks:
The VSAC SVS XML and Excel downloads all use the non-hyphenated versions of the Code System names. I'm curious why Cypress/`health-data-standards` uses any other sort of convention.

eedrummer:
@marks Much of the code in health-data-standards existed before the NLM VSAC service. Originally, it was focused on code for working with the HITSP C32 and ASTM CCR. We created the CodeSystemHelper because HL7 CDA based documents, like the HITSP C32, use OIDs to represent code systems names. Since OIDs are pretty difficult for humans to work with, we translate them to their code system name when they are represented internally. The conventions we use for naming code systems likely reflect what we found in the displayName attribute in code elements in the first C32 documents we were dealing with.

PR #63, opened 2013-10-24
Cleaning up the measure model and associated measure fixtures

eedrummer:
Moving over some code from Cypress that we needed to extract measure
information in the QRDA Category I scoop and filter logic.

Smoking gun logic has been removed. This will be moved to Cypress.

The fixtures for the measures were replaced with the fixtures from
Cypress. These appear to be closer to the JSON structure that is
expected by most of the code for dealing with measures. However,
I'm still finding inconsistencies in these fixtures.


hadleynet:
The new structure of the data criteria field looks odd, why have an array of hashes, where each hash has only a single entry ?

Before:

data_criteria: {
  "data_criteria_id": {// data criteria information}
  ...
}

Now:

data_criteria: [
  {
    "data_criteria_id": {// data criteria information}
  },
  …
]

eedrummer:
@hadleynet I agree that it is odd. However, this is what our measure JSON currently looks like. The HQMF model classes will break if it is switched back to the old form. I only did this to try to get all of the current code in sync with what we produce in our measure bundles. I suspect that this structure is created by our HQMF to JavaScript code. I don't feel qualified to go spelunking there.

@andrequina Do you know why this change was made?

andrequina:
Unless I'm missing something this should not be related to HQMF2JS.  The HQMF Model classes are the code that actually goes into HQMF2JS.

https://github.com/projectcypress/health-data-standards/tree/master/lib/hqmf-model

Bonnie uses a measure model similar to the HQMF Model and passes the HQMF Model into HQMF2JS for javascript generation.

The code being referenced appears to be the measure model that gets exported from bonnie into pop health and cypress for QME.  Changes to this model will only impact Cypress and popHealth as far as I can tell.

Unfortunately this structure was not ideally constructed.  It was initially built to mirror the structure that the 1.0 popHealth was expecting so that the stage 2 measures could be loaded into the legacy popHealth (since at the time popHealth was not being updated).  This initial model was adopted by Cypress because it worked with the existing QME and was extended to add on components needed by Cypress.  I have no idea what the data criteria on this model are used for… if it's used anywhere it would be in popHealth or Cypress, not for javascript generation.

The json for this model is created by bonnie as part of the bundle export process and we can modify that structure based on the needs of popHealth or Cypress.

eedrummer:
@andrequina data criteria are being used in both popHealth and Cypress by QRDA Category I generation. We need it to do scoop and filter. All that the QRDA code needs is the HQMF OID for the data criteria and the value set OID. I don't care too much about how we structure that.

As a side note, the JSON is getting pretty big, so if we have a chance to slim it down, it would be great. As it stands, trying to load the full JSON in a proper Mongoid model takes quite some time (like 70 seconds for all of the EP measures).

hadleynet:
So, would it be possible to update Cypress and PopHealth to use the simple
map rather than the array of 1 element maps ? Keeping the data structure
as simple as possible seems like a worthwhile goal.

Marc.

andrequina:
Yeah, if I recall correctly there's duplication of data as well… in particular I think the measure logic may be duplicated in two different formats.

eedrummer:
Yes. This should be simple. I think all we would need to do is change the all_data_criteria method in the Measure model. I would just need bundles upstream that implement this new structure. Do we want to accept this PR as is and make the changes later, or do we want to change this immediately?

hadleynet:
I'll leave it up to you to decide how best to proceed. The existing PR
seems to move in the opposite direction to that planned and much of it
would need to be reversed later.

Marc.

eedrummer:
Agreed that it would need to be reversed later, however, our current code is out of sync with reality. I'm happy to hold off on this if we can get new bundles quickly. If we feel like new bundles are far off, I'd recommend making this change so that our code matches what is in MongoDB.

hadleynet:
Merging for now with the understanding that we will clean up the measure JSON structure in the not too distant future.

PR #66, opened 2013-11-08
import entry.time (not entry.start_time) for diagnostic study orders

yoon:
Measures that do time comparisons (e.g. 0376) expect an entry #time or both an #start_time and #end_time. Since the diagnostic study order occurs at a single time-point, use the #time value.


rdingwell:
@yoon please resolve failing tests

yoon:
Fixed test

eedrummer:
Rebased and merged

PR #67, opened 2013-11-08
call population_count with 2 arguments

### LINKED ISSUE #61: cat3 continuous variable value template fails on population count
population.population_count('OBSERV') provides only one argument to a method with arity: 2

https://github.com/projectcypress/health-data-standards/blob/master/templates/cat3/_continuous_variable_value.cat3.erb#L8

yoon:
The module method HealthDataStandards::CQM::PopulationSelectors#population_count expects two arguments. Closes #61


rdingwell:
Closing, superseded by rework to the AggregateCount classes and CAT III rendering

PR #68, opened 2013-11-08
ignore blank top level populations in measure section

yoon:
Blank top_level_populations cause cat3 export to fail. This pull request filters out those blank entries without investigating the source.


rdingwell:
The need for this has been superseded by rework to the overall AggregateCount objects.

PR #69, opened 2013-11-08
adding 38 week gestational period

yoon:
Cat1 export is missing the SNOMED-CT code for 38 week gestational period. This will be essential for measure 0469 to pass.


eedrummer:
Rebased and merged in

eedrummer:
I just rebased and merged it. However after a conversation with @rdingwell it sounds like there are some issues with this at the standard level (not a problem with your code). We may have to revisit this.

PR #70, opened 2013-11-08
removal time for medical equipment

yoon:
Some measures (e.g. 0453) expect medical equipment entries to have a removalTime. removalTime is generated as the "high" value in

https://github.com/projectcypress/health-data-standards/blob/master/templates/cat1/_2.16.840.1.113883.10.20.24.3.7.cat1.erb#L14


eedrummer:
Rebased on top of master and merged in.

PR #71, opened 2013-11-09
correct xpath for encounter and reason. add admit_time to encounter

yoon:
The cat1 encounter importer is using a c32 oid, but the export is using a new oid:

https://github.com/projectcypress/health-data-standards/blob/master/templates/cat1/_patient_data.cat1.erb#L4

Some measures (e.g. 0371, 0437, 0438, 0639) require admitTime to be set. admitTime is generated as the "low" value in:

https://github.com/projectcypress/health-data-standards/blob/master/templates/cat1/_2.16.840.1.113883.10.20.24.3.23.cat1.erb#L14

### DIFF
--- a/lib/health-data-standards/import/cda/encounter_importer.rb
+++ b/lib/health-data-standards/import/cda/encounter_importer.rb
@@ -3,10 +3,9 @@ module HealthDataStandards
     module CDA
       # TODO Extract Discharge Disposition
       class EncounterImporter < SectionImporter
-    
-        def initialize(entry_finder=EntryFinder.new("//cda:section[cda:templateId/@root='2.16.840.1.113883.3.88.11.83.127']/cda:entry/cda:encounter"))
+        def initialize(entry_finder=EntryFinder.new("//cda:section[cda:templateId/@root='2.16.840.1.113883.3.88.11.83.127']/cda:entry/cda:encounter or //cda:section[cda:templateId/@root='2.16.840.1.113883.10.20.17.2.4']/cda:entry/cda:encounter"))
           super(entry_finder)
-          @reason_xpath = "./cda:entryRelationship[@typeCode='RSON']/cda:act"
+          @reason_xpath = "./cda:entryRelationship[@typeCode='RSON']/cda:observation"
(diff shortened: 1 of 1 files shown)


rdingwell:
@yoon  This is not correct.  The Cat I importer uses the 2.16.840.1.113883.10.20.24.3.23 oid to lookup encounters.  The initializer in the cda folder has the capability of taken on overridden EntryFinder which the CAT I import functionality makes use of to provide the correct oid.   The cda importers is a starting point for cda documents, most of the other importers make use of the base cda importers but override their functionality for the required spec they are implemented against.

yoon:
@rdingwell Thanks for setting me straight on this one. I'll submit admitTime as a separate pull request.

yoon:
@rdingwell I am still having an issue with the encounter reason. I found a c32 test[1](https://github.com/projectcypress/health-data-standards/blob/master/test/unit/import/c32/encounter_importer_test.rb#L16) that matches this reason xpath:

`@reason_xpath = "./cda:entryRelationship[@typeCode='RSON']/cda:act"`[2](https://github.com/projectcypress/health-data-standards/blob/master/lib/health-data-standards/import/cda/encounter_importer.rb#L9)

but for cat1 the reason xpath seems like it should be:

`@reason_xpath = "./cda:entryRelationship[@typeCode='RSON']/cda:observation"`

Is this a proper place for an "or" in the xpath, or should I be adding cat1 code in a more specific place?

eedrummer:
@yoon I think the best thing to do here is to create an EncouterPerformedImporter class in the Cat I importers. It should subclass the CDA EncounterImporter. You could set the correct XPath for QRDA Cat I in the EncounterPerformedImporter. I'm not sure why the representation of a reason is different in HITSP C32, but I'd prefer to leave our more legacy code alone unless we have a good reason to change it.

Also, I know I just accepted your pull request for admit time, but if you do create an EncounterPerformedImporter, it really makes more sense here, which I think was your original intention.

yoon:
See https://github.com/projectcypress/health-data-standards/pull/83

PR #72, opened 2013-11-09
add ordinality to procedures

yoon:
Some measures (e.g. 0453) require ordinality for procedures.


rdingwell:
@yoon Ordinality for procedures is already incorporated in the procedure_importer for CAT I.  CAT I specific implementation needs are included in the cat I importers and should not be included in the core cda importers

PR #73, opened 2013-11-09
Measure calculation workarounds

yoon:
For those attempting to make measures pass in the short term before the hqmf_utils.js functions in the measure bundles are modified and distributed, these commits provide the values those function require.

See https://groups.google.com/forum/#!topic/pophealth-dev/E4iJJWBfAkc


eedrummer:
@yoon I currently get 1 failure and 10 errors when I try to rebase this code against master. I can't move forward with this PR unless the test suite passes.

eedrummer:
These have been incorporated in pull request #103

PR #74, opened 2013-11-11
add incision_time to procedure imports

yoon:
Some measures (e.g. 0527, 0528) require incision time to be imported with procedures


eedrummer:
@yoon This isn't the correct place to look for procedure incision times. Where you placed the code is for our base CDA importing code which needs to work across HITSP C32, C-CDA, etc. Those specifications do not declare an incision time. 

This is implemented for QRDA Cat I in the procedure performed importer. See here:
https://github.com/projectcypress/health-data-standards/blob/master/lib/health-data-standards/import/cat1/procedure_performed_importer.rb#L28

If there is an issue with the importing of incision times, we should address it there.

yoon:
Updated to use the procedure performed importer mentioned

eedrummer:
Thanks for changing this. It has been rebased and merged in.

PR #75, opened 2013-11-11
aggregate and report measures with sub-measures (via #sub_id)

yoon:
Previously, sub-measures were being ignored. Use the sub_id to retrieve and report on them separately.

### DIFF
--- a/lib/health-data-standards/export/cat_3.rb
+++ b/lib/health-data-standards/export/cat_3.rb
@@ -12,7 +12,12 @@ module HealthDataStandards
       def export(measures, header, effective_date, start_date, end_date, filter=nil,test_id=nil)
         results = {}
         measures.each do |measure|
-          results[measure['hqmf_id']] = HealthDataStandards::CQM::QueryCache.aggregate_measure(measure['hqmf_id'], effective_date, filter, test_id)
+          if measure['sub_id']
+            results[measure['hqmf_id']] ||= {}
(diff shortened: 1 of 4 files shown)


rdingwell:
@yoon This is probably not going to get merged in.  Sub measures are a construct that was created for QME in order to deal with stratifications, multi-numerator and multi populations in the HQMF measures, they do not really exist other than in the QME/HDS internals.   When reporting measures all of the populations and stratifications happen in a single Measure Results section not by the individual population groups.  The aggregate method is there to combine all of the sub measure results into a single object that can then be easily accessed for CAT III generation.  I think there are still some issues with how that combining may be taking place but that is how it is suppose to work.

yoon:
@rdingwell Thanks for your guidance. The issue this pull request attempts to address is that the Cat III file generated doesn't have the numbers or population ids for the stratifications. The main measure passes, but the stratifications fail. I'd like to find the right way of fixing this issue, but I can't see how it would be done without retrieving the proper query cache entries. Should we move this all into the QueryCache.aggregate_measure method?

Any further guidance would be greatly appreciated.

rdingwell:
The aggregate objects model is in need of some work.  I dont think it properly handles things the way it needs to.  I'm pretty sure that multi-numerator and multi-population groups are an issue for it currently and I think it need to be reworked.  

What there should really be is an aggregate model that can contain however many populations as the measure defines.  The population entries on the aggregate object should contain the type (IPP,DENOM,MSRPOPL...) the id  and stratification entries for each stratification mapped by the stratification id.  Each population should also be able to contain as many supplemental data elements as needed as well keeping in mind that supplemental data is not reported for stratifications.  

This would probably result in a few modifications to the CAT III templates as well.

rdingwell:
@yoon this issue should be addressed in the cat_3_stratifications branch.

PR #80, opened 2013-11-30
add admit_time to encounter importer

yoon:
Some measures (e.g. 0371, 0437, 0438, 0639) require admitTime to be set. admitTime is generated as the "low" value in:

https://github.com/projectcypress/health-data-standards/blob/master/templates/cat1/_2.16.840.1.113883.10.20.24.3.23.cat1.erb#L14


eedrummer:
Rebased and merged in.

PR #83, opened 2013-12-10
move reason xpath override and admit_time to Cat1::EncounterPerformedImporter

yoon:
The representation of a encounter reason is at a different xpath for QRDA Category 1 and HITSP C32. To accommodate this behavior, we're creating a cat1 encounter performed importer. This is also a more appropriate location for encounter admit time.

See discussion on: https://github.com/projectcypress/health-data-standards/pull/71


eedrummer:
Rebased and merged

Issue #91, opened 2014-01-16
HealthDataStandards::Import::Cat1::PatientImporter#get_clinical_trial_participant always returns true


yoon:
Since it checks the result of Nokogiri::XML::Node#xpath with .nil?, but #xpath never returns nil, only a blank array. This test should instead check for .blank?

eedrummer:
Fixed by pull request #92

PR #92, opened 2014-01-16
Fix for HealthDataStandards::Import::Cat1::PatientImporter#get_clinical_trial_participantrticipant always returning true. Closes #91

yoon:
HealthDataStandards::Import::Cat1::PatientImporter#get_clinical_trial_participantrticipant always returns true. Fix this by checking .blank? (empty array) instead of .nil? on the nokogiri xpath query


eedrummer:
Rebased and merged

PR #97, opened 2014-01-17
extract values and nested values from procedures

yoon:


eedrummer:
@yoon Why is the value XPath an or expression? It seems to me that in QRDA Cat I, for a Procedure Result, you would need to look in the entryRelationship. Where is value used at the root of the entry?

yoon:
Cat1 templates 2.16.840.1.113883.10.20.24.3.18, 2.16.840.1.113883.10.20.24.3.59, and 2.16.840.1.113883.10.20.24.3.69 use the result_value partial to render a value at the root of the entry.

eedrummer:
Ah... OK. That makes sense. I've rebased this and merged it.

rdingwell:
So is that true for CDA templates as well?  I ask because this is in the cda importer not the catI importer.

PR #98, opened 2014-01-17
cumulative medication duration

yoon:
As discussed here: https://groups.google.com/d/topic/pophealth-dev/1Iw48pRe8cw/discussion

A cumulative medication duration is sometimes provided and could be different than the difference between the start_time attributes of medication entries, or the difference between start_time and end_time attributes in a single medication entry. If the duration is provided, it must have been generated by information that exists outside of the scope of start_time and end_time.

Import a provided cumulative medication duration if provided and override a calculated duration.

### DIFF
--- a/lib/health-data-standards/import/cat1/medication_active_importer.rb
+++ b/lib/health-data-standards/import/cat1/medication_active_importer.rb
@@ -5,11 +5,14 @@ module HealthDataStandards
 
         def initialize
           super(CDA::EntryFinder.new("//cda:substanceAdministration[cda:templateId/@root = '2.16.840.1.113883.10.20.24.3.41']"))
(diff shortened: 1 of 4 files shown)


rdingwell:
@yoon  What I was trying to describe on the email list is that there is no place in CAT I to put the cumulative medication duration. CMD is value that is derived from using the dose , frequency , and refill information or in a hacky way using the start and end dates as those should end up being the same as the calculation.  In addition to that CMD must take into account the total duration minus any gaps.  For instance.  Say a person is on Medication X for 30 days then the go off the medication then later on go back on it for 60 days.  The CMD in this case would be 90 days and would have needed to have been derived from more than a single medication entry. 

I know that we currently have a field for CMD in the ruby models but that may likely be removed in the near future as we rework CMD calculation in the QualityMeasureEngine. Or we would need to supply multiple date ranges in a single medication entry and then derive the value form that.  

Thanks for the hard work but for now I am going to close this pull request without merging it in.

PR #105, opened 2014-02-12
verifying risk category assessment value import

ghost:
modified unit test to assert value import


eedrummer:
Rebased and merged

PR #106, opened 2014-02-18
Fix patient ids when importing by_patient results

sporkd:
This fixes the importer so that the id of the newly created patient record is used instead of the id that existed in the database it was exported from. 

Should resolve issue https://github.com/pophealth/popHealth/issues/106

### DIFF
--- a/lib/health-data-standards/import/bundle/importer.rb
+++ b/lib/health-data-standards/import/bundle/importer.rb
@@ -164,6 +164,14 @@ module HealthDataStandards
             end
 
             contents.each do |document| 
+              if name == "by_patient"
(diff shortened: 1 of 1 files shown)


andrequina:
Accepting.  This should not negatively impact Cypress.  The long term approach should be that popHealth should not load results and should possibly load example patients from Cat 1 rather than the bundle json.

sporkd:
Thanks @andrequina! Good point. I've already gotten pophealth to load patients from cypress generated QRDA I xml. However I'm not clear how to seed the measure definitions without the bundle import script. Do you know where I can find info on how to do that?

PR #109, opened 2014-02-27
Fixing assignment of NPIs to a provider

eedrummer:
A provider should only have a singe NPI. This code makes sure that is
the case.


eedrummer:
Wrong repo... I'll create a new PR from the correct place.

PR #116, opened 2014-03-07
Fixes handling of filters in the query cache

eedrummer:
Previously, this code was adding a 'filter' attribute to the query.
The query_cache actually has a property called 'filters'.
Additionally, this now merges in the filters to allow for properties
such as 'filters.provider_id'.


eedrummer:
Issued from the wrong repo. I'll close this and reopen from the right place.

PR #122, opened 2014-03-21
Fix issues causing test suite to fail.

rbclark:
This contains fixes that should correct all of the things causing failures in the test suite.


rbclark:
Is there anything I can do to help this get merged in? I need it in order to be able to move along with the subsequent pull requests with QME and popHealth.

PR #124, opened 2014-03-28
adding caching of templates to export rendering. Moved ERB creating int...

rdingwell:
...o template helper class


eedrummer:
Rebased and merged

PR #125, opened 2014-03-31
Support category 3 files export for measures with multiple populations

anton-iermolenko:
Per our testing we found that when exporting measure with multiple populations in a single QRDA Category 3 file, it duplicates results of a single population and calculates them wrong.

Following code changes resolve this issue for us.

### DIFF
--- a/lib/health-data-standards/export/cat_3.rb
+++ b/lib/health-data-standards/export/cat_3.rb
@@ -12,7 +12,7 @@ module HealthDataStandards
       def export(measures, header, effective_date, start_date, end_date, filter=nil,test_id=nil)
         results = {}
         measures.each do |measure|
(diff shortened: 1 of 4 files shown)


rdingwell:
Can you please provide examples of what you are describing?  Given the changes you have made I think you may not understand that the aggregate_measure method already aggregates the sub measures into the report so there is no need to add that functionality in the way in which you have implemented it.  

Also, any pull requests that do not pass our continuous integration tests will not be incorporated.

anton-iermolenko:
Looks like it was a bug in v3.2.11 which we based off. I see that latest version iterates over "result.population_groups.each do |pg|" now which should output multiple population groups properly.

Thanks for looking at this. Closing pull request.

PR #131, opened 2014-04-07
Fix display of representedCustodianOrganization in QRDA Cat III

eedrummer:
It appears that the tag_name variable in the _organization template
was not working properly. For the assignedCustodian, the element name
is supposed to be representedCustodianOrganization but it was using
the default of representedOrganization. This change removes the
default value and passes in the tag name with each call to the
partial.


eedrummer:
This was included in #132

PR #132, opened 2014-04-11
Fix Cat3 export and add schema unit test

d-ross:
Adds a cat-3 schema test and fixes extraneous <id/> element. Needs custodian_fix to pass. Future enhancement would be to also run schematron validation.


eedrummer:
Rebased and merged

PR #138, opened 2014-04-28
embedded_in fixes where Mongoid stores Prefilters

eedrummer:
Without this statement, Mongoid was putting Prefilters directly in the measures
collection. This breaks a lot of stuff.


eedrummer:
Travis CI errors were due to a machine time out during the bundle install, not a failing test suite.

PR #139, opened 2014-05-01
Json patient import

rdingwell:


eedrummer:
@rdingwell It looks like File.new breaks on JRuby if you are giving it a directory. I've changed the tests to use Dir instead of File and it seems to work. I think we need to be careful if we use the API in another application that we pass in a Dir instead of a File.

PR #145, opened 2014-05-23
Update experimental mongoid branch to contain latest commits from master

rbclark:
I have my branch rebased in order to contain all commits from master, not really sure how this will work as a pull request since it is obviously not going to merge cleanly, however trying it anyway.


eedrummer:
Merged

PR #154, opened 2014-06-12
decrease memory consumption during import

Xodarap:
The import process loads everything into memory first, then dumps it all into mongo. This exceeds the memory constraints of things like the free EC2 instances, so people might find it hard to use health-data-standards there.

This pull request saves value sets as they come in, lowering memory consumption pretty significantly. After making this change, I was able to get importing to work on a server w/ only 1gb memory.


andrequina:
My guess is that this was done this way originally for performance reasons.  Does your change have a significant impact on the time it takes to load a bundle?

Xodarap:
Yeah, that's what I figured. But no, it seemed just as fast. There are < 1k value sets, so even if it adds a few ms to each call you're only adding a couple of seconds to the build time.

PR #155, opened 2014-06-12
Say if measure is episode_of_care within Measure.categories

carpeliam:
popHealth needs to know on the dashboard whether or not a measure is Episode of Care or not. We're already including whether a measure is Continuous Variable, this adds the ability to check for EoC as well.


eedrummer:
Rebased and merged

PR #158, opened 2014-06-13
Ivl fix

gavin-black:
Fixes the IVL_PQ issue by updating the model as it is parsed

### DIFF
--- a/lib/hqmf-parser/2.0/types.rb
+++ b/lib/hqmf-parser/2.0/types.rb
@@ -60,6 +60,7 @@ module HQMF2
     attr_accessor :low, :high, :width
     
     def initialize(entry, type=nil)
+#, HQFM2::Document::NAMESPACES);
       @type = type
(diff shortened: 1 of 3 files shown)


lib/hqmf-parser/2.0/types.rb
```diff
@@ -77,7 +77,11 @@ def to_model
       lm = low ? low.to_model : nil
       hm = high ? high.to_model : nil
       wm = width ? width.to_model : nil
-      HQMF::Range.new(type, lm, hm, wm)
+      model_type = type
+      if @entry.at_xpath('./cda:uncertainRange', HQMF2::Document::NAMESPACES)
+        model_type = 'IVL_PQ'
+      end
+      HQMF::Range.new(model_type, lm, hm, wm)
```
andrequina, line 84: Let's just make type editable rather than re-creating the range

PR #167, opened 2014-06-25
Updated some rake tasks to use HealthDataStandards::CQM::Bundle instead ...

okeefm:
...of Bundle, which used the wrong model when called from popHealth.


eedrummer:
Rebased and merged

PR #168, opened 2014-06-25
Added provider rollup to JSON record importer

okeefm:
Edited self.import_json in bulk_record_importer.rb to add all ancestor providers to record.provider_performances, such that it mirrors the action of the XML importer.


eedrummer:
Rebased and merged

PR #171, opened 2014-07-01
Add additional information to resolve_provider

rbclark:
This pull request adds support for other applications to access additional data about records assigned to the orphan provider. This will be used by popHealth to log the ID of each record which is assigned to the orphan provider.


rbclark:
Not sure exactly why travis ci is failing on this, I tested on the exact versions that are showing as failed and did not run into the same issues.

okeefm:
I saw this same issue on a couple of my pull requests last week; Andy was able to fix them by restarting the Travis jobs, but we may want to investigate further why it was happening.

okeefm:
Looks like that was the case here; I restarted the offending jobs, and they passed this time.

eedrummer:
Rebased and merged

PR #175, opened 2014-07-21
added support for validating patient import against patient manifest and...

ghost:
... returning list of missing patients


eedrummer:
Rebased and merged

PR #182, opened 2014-08-20
Handle specific occurrence reference to Laboratory Test Result's

jtferns:
### Story
- https://www.pivotaltracker.com/story/show/76617224
### Notes
- set `statusText` to ", Result" if a specific occurrence references a laboratory_test (which is a Result)
### Screenshots
- erroneous three lab test elements
  - screen shot 2014-08-20 at 11 17 23 am
- correct Result and Performed lab test elements
  - screen shot 2014-08-20 at 11 14 19 am


jtferns:
Replacing with #183, which avoids nested ternary operators...

PR #185, opened 2014-08-26
Hqmfr2 template map

ghost:


lib/health-data-standards/util/hqmf_template_helper.rb
```diff
@@ -2,31 +2,49 @@ module HealthDataStandards
...
+
+      def self.definition_for_template_id(template_id, version="r1")
+        template_id_map(version)[template_id]
       end
-      
-      def self.template_id_map
-        if @id_map.blank?
-          template_id_file = File.expand_path('../hqmf_template_oid_map.json', __FILE__)
-          @id_map = JSON.parse(File.read(template_id_file))  
+
```
andrequina, line 9: The case statement for the version seems very redundant... can we pull out the path and have the common code common?

lib/health-data-standards/util/hqmf_template_helper.rb
```diff
@@ -2,31 +2,49 @@ module HealthDataStandards
...
 
-      def self.template_id_by_definition_and_status(definition, status, negation=false)
-        kv_pair = template_id_map.find {|k, v| v['definition'] == definition && 
-                                               v['status'] == status && 
-                                               v['negation'] == negation}
-        if kv_pair
-          kv_pair.first
-        else
-          nil
+      def self.template_id_by_definition_and_status(definition, status, negation=false, version="r1")
```
andrequina, line 24: Again... some common code here that could get pulled out of the case statement

lib/health-data-standards/util/hqmf_template_helper.rb
ghost, line 9: that's fine, i'll make that change

PR #190, opened 2014-09-02
Refactor clinical trial participant import to match current Bonnie model

pkmitre:
After this change clinical trials are added to patient records as conditions, and correctly support date bounds.


pkmitre:
Already merged, github did not mark it as such.

PR #195, opened 2014-09-05
Hqmf template bug

okeefm:
Added a test for, and fix for, a bug that caused the hqmf_template_helper to not work when definitions were requested for both HQMFr1 and HQMFr2 documents in the same context.


okeefm:
I'm not sure how often this version-switching takes place, so I don't know if this method of fixing the issue (switching between versions as requested) makes more sense, or caching both versions of the ID map. The other fix is pretty easy to implement, if that's what would be preferred.

PR #215, opened 2014-10-07
Fix link to HQMF IG

hadleynet:


andrequina:
I noticed the same thing and fixed in a separate pull request on the same day #216

PR #222, opened 2014-10-29
Qrda errata release

ssayer:
Addresses changes introduced by 2014 QRDA Errata Release
- Start/End time for Payer information. Includes empty effectiveTime for cat 3 supplemental data that is now required by the spec.
- Add IDs for reporting period and measure reference templates (ignored on import, they're not actually important).
- Change how measure reference IDs are represented in Cat 1 (root/extension w/OID instead of just root).


eedrummer:
This is a little nit-picky, but do we really need another partial for payer supplemental data? As best as I can tell, the only real difference is the effectiveTime. An if statement in the original partial may prevent duplicate work in the future.

ssayer:
The payer template also has two template IDs, where the others only have one.

eedrummer:
Was talking to @rdingwell about this. You can probably leave effectiveTime in across the board. Having it in for all supplemental data shouldn't hurt anything. Then the only thing that would need to change is that the templateId would need to be refactored so that it could accept an array of ids. Sorry to be a stickler about this, but I'm just concerned about maintaining two copies of the code from the value tag down.

PR #223, opened 2014-11-04
Allow polymorphism in bulk record importer

Xodarap:
A lot of EHRs don't follow standards very strictly. So I want to preprocess their files before sending them off to health data standards; it seems like the best way to do this is to let me create a class like

```
  class MyImporter < HealthDataStandards::Import::BulkRecordImporter
    def self.import(xml_data, provider_map = {})
      doc = self.fix_file(xml_data)
      super("", provider_map, doc)
    end
```

This pull request enables that by replacing internal calls to `BulkRecordImporter.X` with `self.X`.

### DIFF
--- a/lib/health-data-standards/import/bulk_record_importer.rb
+++ b/lib/health-data-standards/import/bulk_record_importer.rb
@@ -7,7 +7,7 @@ module HealthDataStandards
         failed_dir ||= File.join(source_dir, '../', 'failed_imports')
         files = Dir.glob(File.join(source_dir, '*.*'))
         files.each do |file|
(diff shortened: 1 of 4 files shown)


ssayer:
It looks like two of the commits are similar/the same, and the second one just undoes the first one. Could you clean up the commits either by cherry picking the commit you want or squashing these commits?

Xodarap:
Squashed

ssayer:
I think you accidentally pushed up the broken code from #224. I'm going to close this PR, because the branch seems to be messed up. If you'd like to resubmit, please create a new one with the cherry picked/ squashed commits you want in the PR.

PR #224, opened 2014-11-04
Enable updating records

Xodarap:
Currently only demographic attributes like first and last name get updated in `update_or_create`. This is pretty confusing and I don't think it was intended behavior.

This pull request changes that so that entries get updated as well.

### DIFF
--- a/lib/health-data-standards/models/record.rb
+++ b/lib/health-data-standards/models/record.rb
@@ -151,7 +151,6 @@ class Record
       if other_val != []
         my_val = self.send(section) || []
         self.send("#{section}=", my_val.concat(other_val))
-        self.dedup_section!(section) unless my_val == []
       end
     end
   end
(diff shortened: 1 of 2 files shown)


ssayer:
I'm going to summon @eedrummer and @rdingwell to see if there's a reason we do this. I don't think we're using this method in Cypress anywhere (at least a project search reveals no mentions).

ssayer:
Circling back with this. It looks like this caused a failure in the merge test. Please resubmit with a fix (or let me know you're working on it), otherwise I'll close this pull request.

Xodarap:
Is this something that would be added if the merge checks were fixed? I'm not sure what the outcome of your conversation with @eedrummer and @rdingwell was.

ssayer:
After talking to the team, we agreed that this change makes sense. If you can clean up the branch and get the tests passing we'll pull it in.

Issue #40, opened 2013-06-20
Models should be namespaced


pelargir:
This gem defines dozens of models (most of them subclasses of Entry) but fails to namespace them within the HealthDataStandards. This means that my own application's Procedure, Provider, etc. classes get stepped on as soon as I import the gem.

+1 for properly namespacing these models.

dillonwelch:
Created https://github.com/projectcypress/health-data-standards/pull/447 as a compromise attempt at this

PR #228, opened 2014-11-28
Namespaced models.

Wylan:
We ran into a similar issue as in #40. All the classes in the lib/models directory have been namespaced, and all calls to them updated with the namespace.


eedrummer:
Thanks for the PR and sorry for taking so long to get back on this. IMHO, this is a good idea. I intend to merge this in, but it has ripple effects for the projects that depend on it (Bonnie, Cypress, popHealth, etc.)

I'm going to talk with other developers on those projects to develop a strategy for getting this merged in and released.

Wylan:
Sounds good, I figured that there would be some issues merging it because it will cause some breaking changes. I'm glad to have been able to contribute something useful back to the project.

dillonwelch:
:+1: for doing something like this. I'm working on a project using HealthDataStandards and have encountered a naming clash.

Perhaps it's something that could be done in stages? One to a few models at a time?

dillonwelch:
Created https://github.com/projectcypress/health-data-standards/pull/447 as a compromise attempt at this

PR #230, opened 2014-12-01
set preferred code set and reason for provider to provider template

ssayer:
Set preferred code set (SNOMED-CT) and add reason


eedrummer:
Rebased and merged

PR #232, opened 2014-12-03
Removed duplicate reason from communication test update

okeefm:
According to the CDA schema, the `entryRelationship`s should go after the `participant`s, so that's the one I kept.

Also added a reason `entryRelationship` to the communication test fixture.


ssayer:
Maybe I'm looking at this wrong, but it looks like there are still two reason elements in the template?

okeefm:
Oops. Not sure how that didn't get removed. Pushed up another commit.

PR #244, opened 2015-01-07
HTML updates for CMD fulfillment history

okeefm:
Updated the view_helper and _entry template to facilitate arrays of fulfillment histories for CMD. Before it would only display the first medication dispense; now it will display all of them, and input line breaks.


ssayer:
merged

PR #245, opened 2015-01-12
Fixing issue where generic patient characteristic criteria were not bein...

rdingwell:
...g renderd in qrda cat I documents.  Added ability to set a wild card on preferred code set in rendering template to allow any coded set to be used. Added returning on generic patient characteristic template in view helper when an explicit valueset oid mapping is not avaiable.


pkmitre:
This change was merged into bonnie_master, and will be merged into master when the bonnie_master branch is finally merged down.

PR #253, opened 2015-02-06
CDA & QRDA Validators

ssayer:
Validation code from Cypress for Schema and Schematron files. Includes singletons for CDA, QRDA Cat I and QRDA Cat III.

(Code not included in JRuby build because Nokogiri J doesn't handle XSLT very well)


test/unit/validate/xml_file_validator_test.rb
```diff
@@ -0,0 +1,52 @@
...
+       doc = Nokogiri::XML(xml_file)
+       errors = Cat3.instance.validate(doc, {file_name: "filename.xml"})
+       assert errors.empty? , "Should be 0 errors for good cat 3 file"
+    end
+
+      test "should be able to tell when a cat III file is bad due to schematron issues" do
+       xml_file = File.new('test/fixtures/qrda/cat3_bad_schematron.xml')
+       doc = Nokogiri::XML(xml_file)
+       errors = Cat3.instance.validate(doc, {file_name: "filename.xml"})
+       assert 2, "Should be 2 errors for bad cat 3 file"
```
okeefm: Does this assertion actually do anything other than assert that 2 is truthy?

ssayer: My bad, I editted the test incorrectly when I pulled it over from Cypress. It should be counting the number of errors. I'll push an update.

ssayer:
It looks like there may be some issues using this in applications, since it appears to be looking for files from the execution path, not the the gem location. The tests pass since they are run from the project directory.

ssayer:
Force pushed an update to use full paths. @eedrummer and/or @rdingwell please review.

okeefm:
Did the update actually get pushed? I'm not seeing the full paths code.

PR #259, opened 2015-03-24
 allow polymorphism in bulk record importer

Xodarap:
This commit allows people to have polymorphic imports, which helps deal with EHR specific idiosyncrasies. This is a different version of PR #223


ssayer:
rebased and merged

PR #260, opened 2015-03-25
Added indexes to Record and Measure models 

okeefm:
Picked some indexes to add to `Record` and `Measure` models, based on some calls we were making frequently in Cypress.


ssayer:
Rebased and merged

PR #262, opened 2015-03-27
integrate valueset validator

ssayer:
Valueset validator was included in a previous commit but not actually integrated into HDS. Added test from Cypress along with fixtures and refactored it to fit in with HDS validators


okeefm:
Looks like you may have to no-op these tests in JRuby also.

ssayer:
I think it may just be the fact that I no-oped the other tests and so something isn't being included. I think we can adjust it so it works in JRuby, or we can be consistent and just say that the validation package is a no-op in JRuby.

PR #264, opened 2015-04-07
Add bundle download/install tasks to HDS

ssayer:
Tasks for downloading/loading bundles were split between Cypress and HDS. I've moved the Cypress tasks into HDS and tested the results both in Cypress and outside of Rails. All the tasks have been consolidated under the bundle namespace.


rdingwell:
Can you remove the comments in the rake task about cypress, kthanksbye

PR #265, opened 2015-04-09
Validators handle more kinds of input

ssayer:
Validators can handle Files and anything that produces XML with to_s

(also removed a commented out pry statement I found)

### DIFF
--- a/lib/health-data-standards/validate/base_validator.rb
+++ b/lib/health-data-standards/validate/base_validator.rb
@@ -6,8 +6,16 @@ module HealthDataStandards
         ValidationError.new(message: msg, location: loc, file_name: file_name, validator: @name)
       end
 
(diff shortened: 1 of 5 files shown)


lib/health-data-standards/validate/base_validator.rb
```diff
@@ -6,8 +6,17 @@ def build_error(msg, loc, file_name)
         ValidationError.new(message: msg, location: loc, file_name: file_name, validator: @name)
       end
 
-      def get_document(doc)
-        (doc.kind_of? Nokogiri::XML::Document)? doc : Nokogiri::XML(doc.to_s)
+      def get_document(input)
+        # binding.pry if input.is_a? File
```
okeefm: Possibly remove this `binding.pry` statement as well?

PR #266, opened 2015-04-15
Java compatibility for validation

ssayer:
Validators now work in JRuby


lib/health-data-standards/validate/schema_validator.rb
```diff
@@ -13,7 +13,7 @@ def initialize(name, schema_file)
         # Validate the document against the configured schema
         def validate(document,data={})
           doc = get_document(document)
-          @xsd.validate(doc).map do |error|
+          @xsd.validate(doc).map do |error|File
```
okeefm: What does the `File` on this line do?

ssayer: I think it was just a mistype

PR #268, opened 2015-04-21
Additional data validation for cat 1

ssayer:
Combines DataCriteria and ValueSet Validators into a single validator


lib/health-data-standards/validate/validators.rb
```diff
@@ -3,7 +3,8 @@
 require_relative 'schema_validator'
 require_relative 'schematron_validator'
 require_relative 'measure_validator'
-require_relative 'valueset_validator'
+# require_relative 'valueset_validator'
```
okeefm: Should this just be removed, or are we keeping it around in case we need to re-require it?

PR #286, opened 2015-06-11
Updated ValueSet API to V2, and added tests to test the new features/format

okeefm:
The new NLM VSAC API V2 is documented here: http://www.nlm.nih.gov/vsac/support/usingvsac/vsacsvsapiv2.html

Notable changes include:

`effectiveDate` in the `retrieveValueSet` endpoint has been removed in favor of `version`. In order to use `effectiveDate`, the `retrieveMultipleValueSets` endpoint would have to be used, and it has a different XML format than `retrieveValueSet`, and so would require other changes. This API rewrite is designed to be used with `retrieveValueSet`. There is also an API call to get valueset versions for a given OID, but that has not been implemented in Health Data Standards.

The response when no Value Set has been found for an OID has been changed from a status `200` to a `404`. As a result, we now capture the `ResourceNotFound` exception raised by `RestClient`, and throw our own, more descriptive `VSNotFoundError`.

### DIFF
--- a/lib/health-data-standards/util/vs_api.rb
+++ b/lib/health-data-standards/util/vs_api.rb
@@ -2,10 +2,10 @@ require 'rest_client'
 require 'uri'
 module HealthDataStandards
   module Util
-  		class VSNotFoundError < StandardError
-  		end
+  	class VSNotFoundError < StandardError
+  	end
 
-		class VSApi			
+		class VSApi
 			attr_accessor :api_url, :ticket_url, :username, :password
 
 			def initialize(ticket_url, api_url, username, password)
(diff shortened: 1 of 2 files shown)


lib/health-data-standards/util/vs_api.rb
```diff
@@ -2,6 +2,9 @@
 require 'uri'
 module HealthDataStandards
   module Util
+  		class VSNotFoundError < StandardError
+  		end
+
 		class VSApi
```
pkmitre: I'd vote for versioning the API, ie keeping VSApi using the old code, and creating a VSApi2 (or similar) containing the new code.

lib/health-data-standards/util/vs_api.rb
```diff
@@ -12,18 +15,22 @@ def initialize(ticket_url, api_url, username, password)
 				@password = password
 			end
 
-			def get_valueset(oid, effective_date=nil, include_draft=false, &block)
+			def get_valueset(oid, version=nil, include_draft=nil, &block)
```
pkmitre: And, if we do version the API, I'd suggest moving away from optional arguments to an options approach, ie

```
def get_valueset(oid, options = {}, &block)
  options.reverse_merge!(version: nil, include_draft: nil) # Not needed for nil defaults, useful for non-nil
```

That way calls into the api move away from inscrutable code like

```
get_valueset(oid, '5.5', true)
```

to the more easily read

```
get_valueset(oid, version: '5.5', include_draft: true)
```

ssayer: +1

okeefm: Yeah, I'd be fine with this. Either an options hash or keyword arguments, assuming we're only targeting Ruby >=2.0.0.

lib/health-data-standards/util/vs_api.rb
eedrummer: I tend to agree, but with one caveat: only if the VSApi maintains the old version. Why keep around V1 if they don't support it?

okeefm: According to NLM, they're planning on keeping the deprecated V1 around 'until further notice (http://www.nlm.nih.gov/vsac/support/usingvsac/deprecatedvsacapi.html)'.

ssayer: The old version will be supported for "a while" (I don't think we have clear timetable on when), but will eventually be deprecated.

lib/health-data-standards/util/vs_api.rb
ssayer: I think we dropped support for Ruby 2.0 already, we only test on Ruby 2.1, 2.2 and JRuby

lib/health-data-standards/util/vs_api.rb
ssayer: I think that VsApi namespace should be reserved for the current version of the API.

...but that's just like, my opinion, man

lib/health-data-standards/util/vs_api.rb
pkmitre: Bonnie is still on Ruby 2.0 (that's what's on our production server)

ssayer: It probably still works. Travis was unhappy about Ruby 2.0 for some reason, so we had to turn it off.

PR #289, opened 2015-06-18
Removed CCR code/testing code from Health-Data-Standards

okeefm:
As it is no longer used and is breaking in jRuby.


ssayer:
You need to remove the CCR require statements from health-data-standards.rb and there's some mentions of it in the README/documentation as well

okeefm:
Good catch. I removed almost all references to `CCR`, other than those in the changelog.

PR #291, opened 2015-06-19
Validation update

dczulada:
Added Performance Rate Validation.  Refactored QRDA3 "Reported Results" extraction from Cypress code into HDS.


ssayer:
Can you change it so that the validators don't assign to instance variables (@errors, @document). I don't think that it would cause an issue since they're reassigned every time, but the convention is to just use a local variable.

PR #296, opened 2015-06-28
HQMFv2 parser fixes and tests

okeefm:
- Fixed the name of the 'parse_fields' HQMFv2 parser method, as before it was 'parse_feilds'
- Changed the xpath of the setId statement to '@root' from '@extension', as apparently that's how HQMFv2 works
- Added tests for the parse_fields methods for both HQMFv1 and v2 to the parser unit test


ssayer:
Bonnie master was merged into master

PR #299, opened 2015-07-08
Performance rate returns nullFlavor NA when Denominator is 0

dczulada:


okeefm:
So it appears that the schematrons require that performance rate provide an `xsi:type` of `REAL` even when `nullFlavor` is provided instead of value.

templates/cat3/_performance_rate.cat3.erb
```diff
@@ -4,7 +4,11 @@
     displayName="Performance Rate" 
     codeSystemName="2.16.840.1.113883.6.1"/>
   <statusCode code="completed"/>
+  <% if population_group.denominator_count - population_group.denominator_exclusions_count - population_group.denominator_exceptions_count > 0 %>
```
ssayer: Can you move this into a helper function?

PR #300, opened 2015-07-08
Cherry-picked commits from bonnie-master regarding fulfills

dczulada:


okeefm:
This probably didn't need to be closed; there's still a test running. We'll see if it finishes.

PR #306, opened 2015-07-16
qrda r3 validation and export

ssayer:
Importer/Templates support QRDA R3, R2 supported under CatIR2 namespace


Gemfile.lock
```diff
@@ -17,7 +17,7 @@ PATH
       uuid (~> 2.3.7)
 
 GEM
-  remote: https://rubygems.org/
+  remote: http://rubygems.org/
```
okeefm: Was this change supposed to be committed?

PR #311, opened 2015-07-21
HDS 3.6.0 release

ssayer:


okeefm:
I think you need to run `bundle install`, and commit the updated `Gemfile.lock`, because Bundler on Travis is looking for HDS 3.5.3, which doesn't exist since we bumped the version to 3.6.0. Running the install updates the version number in the lock file.

PR #316, opened 2015-08-18
Return performance rate validation to not rounding expected values

okeefm:
Problem: Performance Rates (according to the spec) should NOT be rounded to 6 decimal places, but rather truncated. HDS's Performance Rate Valdiator was changed to round them in a previous pull request.

Solution: Change the code back to the way it was before that pull request, and add comments to the performance rate valdiator talking about why we're not rounding the performance rate validator. Also, add tests to ensure that we're correctly validating repeating decimals such as 2/3


dczulada:
Turns out the IG was changed to use rounding for decimal greater than 6 digits.  The validation is actually correct, the error message is the piece that is wrong.  The error message is not displaying the rounded value.

PR #318, opened 2015-09-08
NLM authentication

dczulada:


ssayer:
It looks like this is actually calling out over the network when it does the test. I think it's better to use a WebMock response with this to do the test.

Also there should be a positive and a negative test.

okeefm:
We do some WebMock stuff with testing the ValueSet API, which is similar, in `test/unit/util/vs_api_test.rb`.

ssayer:
:+1:

ssayer:
Rebased and merged into master

PR #319, opened 2015-09-08
Added an always-failing test

okeefm:


okeefm:
Don't merge me, this is for test purposes only.

PR #323, opened 2015-10-12
Xml export

edeyoung:
Bug fix for https://jira.oncprojectracking.org/browse/BONNIE-100.

The reason template only got processed for negation reasons, not for all reasons. According to the QRDA guidance, the template should be used for all reasons. Modified the template to work for both `item.reason` and `item.negation_reason`.

Additionally, found that the Procedure Physical Exam template was not exporting associated reasons. According to the QRDA documentation, it should. Added reason to the template.

### DIFF
--- a/templates/cat1/_2.16.840.1.113883.10.20.24.3.59.cat1.erb
+++ b/templates/cat1/_2.16.840.1.113883.10.20.24.3.59.cat1.erb
@@ -1,6 +1,6 @@
-<% vals = entry.values.empty? ? [nil] : entry.values 
-   vals.each do |value| 
- %>   
+<% vals = entry.values.empty? ? [nil] : entry.values
+   vals.each do |value|
+ %>
 <entry>
   <observation classCode="OBS" moodCode="EVN" <%== negation_indicator(entry) %>>
     <!-- Procedure Activity Procedure (Consolidation) template -->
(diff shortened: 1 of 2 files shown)


templates/cat1/_2.16.840.1.113883.10.20.24.3.59.cat1.erb
```diff
@@ -1,6 +1,6 @@
-<% vals = entry.values.empty? ? [nil] : entry.values 
-   vals.each do |value| 
- %>   
+<% vals = entry.values.empty? ? [nil] : entry.values
+   vals.each do |value|
+ %>
```
pkmitre, line 3: Looks like this might be whitespace-related diffs related to an editor deleting trailing whitespace in a file on save... it would be useful to tweak that in your editor settings, not an issue here but sometimes it creates lots of non-essential diffs that can be difficult to wade through.

templates/cat1/_reason.cat1.erb
```diff
@@ -1,22 +1,31 @@
-<% if entry.negation_reason.present? 
-      vset = oid_for_code(entry.negation_reason,reason_oids, entry.record["bundle_id"])
+<%  if entry.negation_reason.present?
+      reason = entry.negation_reason
+    elsif entry.reason.present?
+      reason = entry.reason
+    end
+    
+    if reason
```
pkmitre: The original code uses the 'present?' method, which is subtly different from a simple truthiness test in that it will return false for an empty string (and empty array or hash) where the truthiness test will return true (in ruby, only false and nil are falsy). I'm not sure the distinction is important here (seems unlikely that there are patient records with an empty string for reason), but then again I don't know why present? was used to begin with...

PR #329, opened 2015-11-04
Made _reason template work with both objects and hashes

edeyoung:
JIRA: https://jira.oncprojectracking.org/browse/BONNIE-113

This was caused by an `Entry` object being passed into the `_reason.cat1.erb` template. `_reason.cat1.erb` expects a hash (it calls the `has_key?` method). Made `_reason.cat1.erb` able to handle both objects and hashes.

### DIFF
--- a/templates/cat1/_reason.cat1.erb
+++ b/templates/cat1/_reason.cat1.erb
@@ -20,7 +20,7 @@
     <value xsi:type="CD"
          code="<%= reason['code'] %>"
          codeSystem="<%= HealthDataStandards::Util::CodeSystemHelper.oid_for_code_system(reason['codeSystem'] || reason['code_system']) %>"
-       <% if reason.has_key?('title') -%>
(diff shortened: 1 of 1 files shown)


templates/cat1/_reason.cat1.erb
```diff
@@ -20,7 +20,7 @@
     <value xsi:type="CD"
          code="<%= reason['code'] %>"
          codeSystem="<%= HealthDataStandards::Util::CodeSystemHelper.oid_for_code_system(reason['codeSystem'] || reason['code_system']) %>"
-       <% if reason.has_key?('title') -%>
+       <% if reason.respond_to?('title') || reason.has_key?('title') -%>
          displayName="<%=reason['title']%>"
```
pkmitre, line 24: Does reason['title'] still work when reason is an object? Just want to make sure, since the respond_to? test is testing for the presence of reason.title().

edeyoung, line 24: Yes, still works

PR #330, opened 2015-11-06
HQMF r2.1 parser fix

elehner:
Fixed HQMF r2.1 parsing


lib/hqmf-model/data_criteria.json
```diff
@@ -115,7 +115,7 @@
       "title":"patient characteristic",
       "category":"characteristic",
       "definition":"patient_characteristic",
-      "status":"",
+      "status":"clinical_trial_participant",
```
edeyoung: From chatting with Eric, it sounds like this change was necessary because the SimpleXML parser includes this `status` for patient characteristic clinical trial participant. I'm wondering if we should change both this and the SimpleXML parser to use `sub_category`. 

Looking at the docs (QDM 4.2), it looks like "clinical trial participant" is specified as part of the datatype. It looks like datatypes are included either in the `status` field or in the `sub_category` field within data_criteria.json. However, `status` seems to be used for action-oriented things like "ordered" and "performed". `sub_category` seems to be used for descriptors like "intollerance" and "allergy". `sub_category` is also used for other patient characteristics like "gender" and "birthdate". "clinical trial participant" seems to align better with the use of `sub_category` than with the use of `status`.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -163,36 +79,36 @@ def handle_variable_subsets
...
       # TODO: Figure out when exactly to handle variable references as verbose
-      if reference && references.length==1
+      if reference && references.length == 1
         ref_id = strip_tokens(HQMF2::Utilities.attr_val(reference, 'cda:id/@extension'))
         verbose_ref_id = strip_tokens("#{HQMF2::Utilities.attr_val(reference, 'cda:id/@extension')}_#{HQMF2::Utilities.attr_val(reference, 'cda:id/@root')}")
         reference_criteria = @data_criteria_references[ref_id] if ref_id
         reference_criteria = @data_criteria_references[verbose_ref_id] if verbose_ref_id && !reference_criteria
         unless reference_criteria
-          puts "MISSING_DC_REF: #{ref_id} & #{verbose_ref_id}"
+          #puts "MISSING_DC_REF: #{ref_id} & #{verbose_ref_id}"
```
edeyoung: consider commenting or deleting entire `unless` block. A delete seems appropriate here.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -163,36 +79,36 @@ def handle_variable_subsets
...
+        end
+        if isGrouper && reference_criteria.try(:variable)
+          idExtension_xpath = './*/cda:id/@extension'
+          idRoot_xpath = './*/cda:id/@root'
+          return if !(attr_val(idExtension_xpath) =~ /^occ[A-Z]of_qdm_var_/).nil?
+          @verbose_reference = true
+          #puts "Updated grouper: #{@id} #{@verbose_reference}"
```
edeyoung: consider deleting commented debug statement.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -303,6 +221,20 @@ def extract_type_from_template_id
...
+              # apply information from settings to the data criteria
+              @title = settings["title"] if !settings["title"].blank?
+              @definition = settings["definition"] if !settings["definition"].blank?
+              @status = settings["status"] if !settings["status"].blank?
+              @category = settings["category"] if !settings["category"].blank?
+              @hard_status = settings["hard_status"]
+            end
+          rescue Exception
+            # ignore error (none should exist)
+          end unless @status
```
edeyoung: the structure of `end unless @status` is confusing to me. What about putting the `unless` at the beginning instead?

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -303,6 +221,20 @@ def extract_type_from_template_id
         if defs
           @definition = defs['definition']
           @status = defs['status'].length > 0 ? defs['status'] : nil
+          # Only attempt to grab settings if no status is present
+          begin
+            settings = HQMF::DataCriteria.get_settings_for_definition(defs['definition'], defs['status'])
```
edeyoung: looking at `get_settings_for_definition`, it looks like it raises an error if there is no setting for that definition (lib/hqmf-model/data_criteria.rb). given that, consider removing the `if` statement and adding code into the `rescue` statement (or a comment stating that you do nothing if there's no settings).

edeyoung: consider modifying `get_settings_for_definition(defs['definition'], defs['status'])` to `get_settings_for_definition(@definition, @status)`

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -303,6 +221,20 @@ def extract_type_from_template_id
         if defs
           @definition = defs['definition']
           @status = defs['status'].length > 0 ? defs['status'] : nil
+          # Only attempt to grab settings if no status is present
```
edeyoung: more out of curiosity than anything, but why?

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -423,15 +362,46 @@ def to_model
...
+    # def extract_source_data_criteria
+    #   DataCriteria.new(@entry, @data_criteria_references, @occurrences_map).extract_as_source_data_criteria(@id, @source_data_criteria)
+    # end
 
     # Return a new DataCriteria instance with only grouper attributes set
     def extract_variable_grouper
       return unless @variable
+      if @do_not_group
+        if !@data_criteria_references["GROUP_#{@children_criteria.first}"].nil?
+          @children_criteria[0] = "GROUP_#{@children_criteria.first}" if @children_criteria && @children_criteria.length == 1
```
edeyoung: looks like a check is being performed for existence of `@children_criteria` after `@children_criteria.first` is referenced in the preceding `if` statement. If `!!@children_criteria == false`, then an error would be thrown before the check happens. Consider removing the check if you know `!!@children_criteria == true` or move the check earlier.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -423,15 +362,46 @@ def to_model
...
+        elsif @children_criteria.length == 1
+          reference_criteria = @data_criteria_references[@children_criteria.first] if @children_criteria.first
```
edeyoung: if `!!@children_criteria.first == false`, then `reference_criteria` is `nil`. This will create exceptions with the code below. consider adding a existence check (`unless reference_criteria.nil?`) before the code below is executed.

lib/hqmf-parser/2.0/data_criteria.rb
elehner: There was an error popping up during the test when statuses were included. From what I can make out, it attempts to use the to_model function, and fails due to the use of `get_settings_for_definition` inside of the initialize function in `lib/hqmf-model/data_criteria.rb`.

lib/hqmf-parser/2.0/data_criteria.rb
BuffaloCoder: Good point. At the time, the rescue was not capturing the failure properly, and I meant to ask about it, but the unless was a bandaid. I'll try to remove the unless and get it to work. Is there anything obvious I'm missing with the rescue that you can see that would make it fail anyway?

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -423,15 +362,46 @@ def to_model
...
+        @type = reference_criteria.subset_operators
+        @definition = reference_criteria.definition
+        @status = reference_criteria.status
+        @code_list_id = reference_criteria.code_list_id
+        @temporal_references = reference_criteria.temporal_references
+        @subset_operators = reference_criteria.subset_operators
+        @variable = reference_criteria.variable
+        @value = reference_criteria.value
+        @children_criteria = []
+      end
```
edeyoung, line 362: Looks like there's a good chunk of repeated code in the `if @do_not_group` clause and the following clause. Is there a way to rework the code so this is only done once?

lib/hqmf-parser/2.0/data_criteria.rb
elehner: Since there is only one clause/block in an if/unless statement, I wanted to inline form of unless (cuts down on the two lines for separate unless and end).  We can make a practice out of placing blocks inside dedicated if/unless blocks if that's preferable.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -303,6 +221,20 @@ def extract_type_from_template_id
...
+            if settings
+              # apply information from settings to the data criteria
+              @title = settings["title"] if !settings["title"].blank?
+              @definition = settings["definition"] if !settings["definition"].blank?
+              @status = settings["status"] if !settings["status"].blank?
+              @category = settings["category"] if !settings["category"].blank?
+              @hard_status = settings["hard_status"]
+            end
+          rescue Exception
+            # ignore error (none should exist)
```
pkmitre: If we don't expect an error, then why the rescue? In cases like that I'd rather leave out the rescue and, if there is at some point an error, we'll find out about it, which I think is preferable to it being silently ignored.

PS I agree about the end unless @status being confusing, when reading the code linearly it can easily be missed. A good rule of thumb is to only use post-conditionals on fairly short simple lines.

elehner: The rescue was originally an attempt to solve the issue that was actually solved by the "unless @status". I'll fix the code as noted and see if I can't constrain the "unless" to better handle the specific case I was encountering.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -423,43 +362,73 @@ def to_model
...
+    # end
 
     # Return a new DataCriteria instance with only grouper attributes set
     def extract_variable_grouper
       return unless @variable
+      if @do_not_group
+        if !@data_criteria_references["GROUP_#{@children_criteria.first}"].nil?
+          @children_criteria[0] = "GROUP_#{@children_criteria.first}" if @children_criteria.length == 1
+        elsif @children_criteria.length == 1
+          return if @children_criteria.first.blank?
```
edeyoung: consider adding as an `&&` to the `if` statement above

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -423,43 +362,73 @@ def to_model
...
+          reference_criteria = @data_criteria_references[@children_criteria.first]
+          duplicate_child_info(reference_criteria)
+          @children_criteria = reference_criteria.children_criteria
+        end
+        return
+      end
       @variable = false
       @id = "GROUP_#{@id}"
+      if @children_criteria.length == 1 && @children_criteria[0] =~ /GROUP_/
+        return if @children_criteria.first.blank?
```
edeyoung: does `@children_criteria[0] =~ /GROUP_/` in a sense perform the `@children_criteria.first.blank?` check for you? can you delete this line?

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -589,11 +558,12 @@ def extract_negation
     def extract_child_criteria
       @entry.xpath("./*/cda:outboundRelationship[@typeCode='COMP']/cda:criteriaReference/cda:id", HQMF2::Document::NAMESPACES).collect do |ref|
         child_ref = Reference.new(ref)
-        unless @data_criteria_references.keys.include?(strip_tokens child_ref.id)
+        if @data_criteria_references.keys.include?(strip_tokens child_ref.id)
```
edeyoung: out of curiosity again, but what's the reason for the reversal in logic?

lib/hqmf-parser/2.0/data_criteria.rb
pkmitre: FYI the opposite of blank? is present? (you'll need to flip the sense if you use this suggestion, I think)

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -589,11 +558,12 @@ def extract_negation
...
           # puts "Updated CC: #{child_ref.id}"
           child_ref.update_verbose(true)
-          puts "ERROR\t Could not find verbose CC: #{child_ref.id}" unless @data_criteria_references.keys.include?(strip_tokens child_ref.id)
+          # puts "ERROR\t Could not find verbose CC: #{child_ref.id}" unless @data_criteria_references.keys.include?(strip_tokens child_ref.id)
```
edeyoung: consider deleting.

speaking of error/debug statements, do we have a logger for all of this code? if not, should we make one?

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -589,11 +558,12 @@ def extract_negation
...
         end
+        # TODO  This may not be correct. This makes the child-criteria refer to the source, rather than it's own data criteria.
```
edeyoung: thoughts on how we should validate this?

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -730,9 +624,20 @@ def extract_specific_or_source
...
-      if @definition == 'derived' && @children_criteria.empty?
-        # puts "Fixing SO grouper empty children for #{@id} with #{@source_data_criteria}"
-        @children_criteria << @source_data_criteria
+      if @definition == 'derived'
+        if @children_criteria.empty?
+          # puts "Fixing SO grouper empty children for #{@id} with #{@source_data_criteria}"
+          @children_criteria << @source_data_criteria
+        end
+        if @children_criteria.length == 1 && (@children_criteria.first == @source_data_criteria || @source_data_criteria.nil?)
+          reference_criteria = @data_criteria_references[@children_criteria.first] if @children_criteria.first
```
edeyoung: can the `if @children_criteria.first` check be a part of the `if` statement above? maybe replacing the `@children_criteria.length == 1` check?

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -742,8 +647,10 @@ def extract_field_values
       @entry.xpath('./*/cda:outboundRelationship[*/cda:code]', HQMF2::Document::NAMESPACES).each do |field|
         code = HQMF2::Utilities.attr_val(field, './*/cda:code/@code')
         code_id = HQMF::DataCriteria::VALUE_FIELDS[code]
-        unless @negation && code_id == "REASON"
+        # No need to run if there is no code id
+        unless (@negation && code_id == "REASON") || code_id.nil?
           value = DataCriteria.parse_value(field, './*/cda:value')
+          value ||= DataCriteria.parse_value(field, './*/cda:effectiveTime')
           fields[code_id] = value if value && code_id
```
edeyoung: we can get rid of `&& code_id` now with the addition of `|| code_id.nil?` to the `unless` statement?

lib/hqmf-parser/2.0/document.rb
```diff
@@ -120,19 +69,26 @@ def initialize(hqmf_contents)
...
+        end
         handle_variable(criteria) if criteria.variable
-      end
 
-      # Extract the source data criteria from data criteria
-      @source_data_criteria = @data_criteria.map{|dc| dc.extract_source_data_criteria}.uniq { |sdc| sdc.id }
+        @reference_ids.concat(criteria.children_criteria)
+        criteria.temporal_references.each do |tr|
+          @reference_ids << tr.reference.id if tr.reference.id != HQMF::Document::MEASURE_PERIOD_ID
+        end if criteria.temporal_references
```
edeyoung: consider adding `if` statement to top of loop for readability

lib/hqmf-parser/2.0/document.rb
```diff
@@ -221,7 +178,15 @@ def initialize(hqmf_contents)
...
       @populations.push *@stratifications
-      handle_verbose_references
+      
+      @reference_ids.uniq
+
+      # Remove any data criteria from the main data criteria list that already has an equivalent member and no references to it
+      # The goal of this is to remove any data criteria that should not be purely a source
+      # Ignore any referenced by child criteria
+      base_criteria_defs = ['patient_characteristic_ethnicity', 'patient_characteristic_gender', 'patient_characteristic_payer', 'patient_characteristic_race']
+      @data_criteria.reject! {|dc| @reference_ids.index(dc.id).nil? && !base_criteria_defs.include?(dc.definition) && !@data_criteria.detect{|dc2| dc != dc2 && dc.code_list_id == dc2.code_list_id && detect_criteria_covered_by_another(dc, dc2)}.nil?}
```
edeyoung: Consider breaking this logic into several lines. There's a lot going on here, and I think readability would improve if it were broken up.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -366,6 +336,29 @@ def find(collection, attribute, value)
...
+    # Check if one data criteria contains the others information by checking that one has everything the other has (or more)
+    def detect_criteria_covered_by_another(data_criteria, check_criteria)
+      same_definition = data_criteria.definition == check_criteria.definition
+      same_status = data_criteria.status == check_criteria.status
+      same_children = data_criteria.children_criteria.sort.join(",") == check_criteria.children_criteria.sort.join(",")
+      derived_variable = data_criteria.variable
+      derived_operator = data_criteria.derivation_operator
+      derived_subset_operator = !data_criteria.subset_operators.empty?
+
+      same_value = data_criteria.value.nil? && !check_criteria.value.nil? || data_criteria.value.try(:to_model).try(:to_json) == check_criteria.value.try(:to_model).try(:to_json)
```
edeyoung: should `!check_criteria.value.nil?` be `check_criteria.value.nil?` - no bang?

edeyoung: Also, my personal preference is to use parens when combining ands and ors. This is correct as is, but I find using parens helps me with readability and also makes it so I don't have to worry about order of operations.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -366,6 +336,29 @@ def find(collection, attribute, value)
...
+      same_temporal_references = check_criteria.temporal_references.nil? || data_criteria.temporal_references.nil? && !check_criteria.temporal_references.nil? || data_criteria.temporal_references.empty?
```
edeyoung: this is confusing to me. wouldn't i want both the `check_criteria` and the `data_criteria` to both be nil if one is nil?

Also, same comment regarding parens and mixing ands and ors.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -366,6 +336,29 @@ def find(collection, attribute, value)
...
+      same_field_values = data_criteria.field_values.nil? && !check_criteria.field_values.nil? || data_criteria.field_values.try(:to_json) == check_criteria.field_values.try(:to_json) || data_criteria.field_values.empty?
+      same_negation_values = data_criteria.negation_code_list_id.nil? && !check_criteria.negation_code_list_id.nil? || data_criteria.negation_code_list_id == check_criteria.negation_code_list_id
```
edeyoung: same comments for `same_field_values` and `same_negation_values`.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -424,58 +417,80 @@ def detect_missing_specifics
...
-      ref_prcns = []
-
-      @population_criteria.each do |criteria|
-        criteria.preconditions.each do |c_prcn|
-          extract_preconditions(c_prcn, ref_prcns)
-          ref_prcns << c_prcn if c_prcn.reference
+    def extract_preconditions(precondition, list)
+      unless precondition.preconditions.empty?
+        precondition.preconditions.each do |prcn|
+          extract_preconditions prcn, list
```
edeyoung: consider putting arguments in parens to match styling of the rest of the code

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -122,7 +122,7 @@ def self.parse_date_fields(entry,fields)
         fields[field] = Range.new(low, 'IVL_PQ')
       end
       if high
-        field = "STOP_DATETIME"
+        field = "DISCHARGE_DATETIME"
         if template_ids.include?('2.16.840.1.113883.10.20.28.3.26')
```
edeyoung: do we need to add a `template_ids.include?` thing here for when `STOP_DATETIME` might be appropriate (like how below it has one for `DISCHARGE_DATETIME` which is now the default)?

Same with `START_DATETIME` above.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -424,58 +426,80 @@ def detect_missing_specifics
...
-      criteria_ids = @data_criteria.map(&:id)
-      referenced_criteria_ids = []
-      ref_prcns = []
-
-      @population_criteria.each do |criteria|
-        criteria.preconditions.each do |c_prcn|
-          extract_preconditions(c_prcn, ref_prcns)
-          ref_prcns << c_prcn if c_prcn.reference
+    def extract_preconditions(precondition, list)
+      unless precondition.preconditions.empty?
```
edeyoung: looks like this can be removed - would just be iterating over an empty list so nothing would happen. consider removing.

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -111,21 +111,17 @@ def self.parse_date_fields(entry,fields)
       low = entry.at_xpath("./cda:effectiveTime/cda:low/..")
       high = entry.at_xpath("./cda:effectiveTime/cda:high/..")
       if low
-        field = "START_DATETIME"
+        field = "ADMISSION_DATETIME"
```
edeyoung, line 114: should we add in a `template_ids` check for `START_DATETIME`?
Same for `STOP_DATETIME` below.

lib/hqmf-parser/2.0/precondition.rb
```diff
@@ -14,29 +14,38 @@ def self.parse(entry, doc, id_generator)
...
       if aggregation
         precondition_entries = entry.xpath('./*/cda:precondition', HQMF2::Document::NAMESPACES)
         preconditions = precondition_entries.collect do |precondition|
-          Precondition.parse(precondition, doc, id_generator)
+          precondition = Precondition.parse(precondition, doc, id_generator)
+          if precondition.reference.nil? && precondition.preconditions.empty?
+            nil
+          else
+            precondition
+          end
```
edeyoung, line 34: should this chunk of `precondition` code be made into a helper method? It's used here and in `population_criteria.rb`

lib/hqmf-parser/2.0/precondition.rb
```diff
@@ -14,29 +14,38 @@ def self.parse(entry, doc, id_generator)
...
         conjunction = aggregation.name
         case conjunction
         when "allFalse"
           negation = true
           conjunction = "atLeastOneTrue"
+          return self.new(id_generator.next_id,"atLeastOneTrue",[self.new(id_generator.next_id,conjunction,preconditions,negation,reference)])
         when "atLeastOneFalse"
           negation = true
           conjunction = "allTrue"
+          return self.new(id_generator.next_id,"allTrue",[self.new(id_generator.next_id,conjunction,preconditions,negation,reference)])
```
edeyoung, line 46: consider expanding this and line 42 into multiple lines for readability.

lib/hqmf-parser/2.0/field_value_helper.rb
elehner, line 114: I haven't found a template ID that uses START/STOP_DATETIME.  Are there specific ids that require it?

lib/hqmf-parser/2.0/types.rb
```diff
@@ -37,28 +37,44 @@ def initialize(entry, default_type='PQ', force_inclusive=false)
...
       greater_than_equal_tr = attr_val("../cda:high/@nullFlavor")=="PINF" &&
         attr_val("../cda:low/@value") &&
         attr_val("../@lowClosed")=='true'
-      equivalent = attr_val("../@lowClosed")=='true' &&
-        attr_val("../@highClosed")=='true' &&
-        attr_val("../cda:low/@value")==attr_val("../cda:high/@value")
+
+      equivalent = attr_val("../cda:low/@value")==attr_val("../cda:high/@value")
+        attr_val("../@lowClosed") !='false' &&
+        attr_val("../@highClosed") !='false'
```
edeyoung, line 50: it looks like a logical operator might be missing at line 48.

Also, why `!= 'false'` instead of `=='true'`?

lib/hqmf-parser/2.0/types.rb
```diff
@@ -37,28 +37,44 @@ def initialize(entry, default_type='PQ', force_inclusive=false)
...
+      #  a more solid solution can be reached
       # subset - EP128, EH108
       greater_than_equal_ss = attr_val("../cda:low/@value")!="0" &&
         !attr_val("../cda:high/@value") &&
-        attr_val("../@lowClosed")!='false'
+        attr_val("../@lowClosed")!='false' &&
+        !attr_val("../../../../../qdm:subsetCode/@code").nil?
 
-      less_than_equal_tr || less_than_equal_los || greater_than_equal_tr || greater_than_equal_ss || equivalent || @force_inclusive
+      less_than_equal_los || less_than_equal_tr || greater_than_equal_los || greater_than_equal_ss || greater_than_equal_tr || greater_than_equal_v || equivalent || @force_inclusive # || greater_than_equal_ss
```
edeyoung: consider removing commented out trailing clause

lib/hqmf-parser/2.0/types.rb
elehner, line 50: Right, there should be.  One the `!= 'false'`, there are some cases we've seen where the SimpleXML marks the values as inclusive, but there doesn't exist an instance of lowClosed or highClosed, so if they are either marked true or are non-existent, then they can be inclusive.

lib/hqmf-parser/2.0/precondition.rb
elehner, line 34: So in the refactor, I was planning on reducing it down to `precondition.reference.nil? && precondition.preconditions.empty? ? nil : precondition`, and the upper line.  Since this is only used in 2 places, I don't think a helper method would be especially useful.

lib/hqmf-parser/2.0/precondition.rb
elehner, line 46: I changed it a bit here on the refactor branch: https://github.com/projectcypress/health-data-standards/blob/hqmf_r2.1_refactor/lib/hqmf-parser/2.0/precondition.rb#L48. Should we change it here or wait until the refactor?

PR #332, opened 2015-11-23
HQMF 2.1 refactor

elehner:
Refactors the HQMF 2.0 parser and the HQMF vs Simple Test to improve readability and understanding


lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -28,227 +18,28 @@ class DataCriteria
...
-      # Need to handle grouper criteria that do not have template ids -- these will be union of and intersection criteria
-      if @template_ids.empty?
-        # Change the XPRODUCT to an INTERSECT otherwise leave it as a UNION
-        @derivation_operator = HQMF::DataCriteria::INTERSECT if @derivation_operator == HQMF::DataCriteria::XPRODUCT
-        @description ||= (@derivation_operator == HQMF::DataCriteria::INTERSECT) ? "Intersect" : "Union"
-      end
+      obtain_specific_and_source = SpecificOccurrenceAndSource.new(@entry, @id, @local_variable_name, @data_criteria_references, @occurrences_map)
+      # Pulling these 5 variables out via destructing
+      @source_data_criteria, @source_data_criteria_root, @source_data_criteria_extension,
+        @specific_occurrence, @specific_occurrence_const = obtain_specific_and_source.extract_specific_and_source
```
edeyoung: can the source data criteria extraction and the specific occurrence extraction be done separately? If so, consider splitting into two different function calls.

Additionally, consider renaming the function to something like `extract_specific_occurrences` and `extract_source_data_criteria` or if it needs to be kept together (and I know this is long, but for some reason as is it reads a bit confusingly for me) `extract_specific_occurrences_and_source_data_criteria`.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -271,64 +62,20 @@ def code_list_id
...
-      codeValue = attr_val("#{@code_list_xpath}/@code")
-      if codeSystemName && codeValue
-        {codeSystemName => [codeValue]}
-      else
-        nil
-      end
-    end
-
     def to_model
+      mv = value.try(:to_model)
```
edeyoung, line 71: consider adding a comment to this method.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -340,14 +87,7 @@ def to_model
     def extract_variable_grouper
```
edeyoung, line 96: maybe this isn't the most appropriate place for it, but it would be nice to have whatever `grouper` attributes are defined or explained somewhere. Something like this could point to the definition.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -365,18 +105,6 @@ def extract_variable_grouper
...
-      @code_list_id ||= child_ref.code_list_id
-      @temporal_references = child_ref.temporal_references if @temporal_references.empty?
-      @subset_operators ||= child_ref.subset_operators
-      @variable ||= child_ref.variable
-      @value ||= child_ref.value
-    end
-
     # Set this data criteria's attributes for extraction as a grouper data criteria
     # for encapsulating a variable data criteria
     # SHOULD only be called on the variable data criteria instance
```
edeyoung: This comment doesn't make a whole lot of sense to me. Consider rewording. 

Things that aren't clear to me include:
- what is meant by "attributes for extraction"?
- what is a grouper data criteria (see comment above about doing a general definition)
- what is meant by "for encapsulating a variable data criteria"? Does this mean that whatever this data criteria is, it will include a variable data criteria in it or a reference to a variable data criteria?

Is it worth explaining the different kinds of data criteria or is this something that is expected for the developer to already know?

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -340,14 +87,7 @@ def to_model
...
-        if !@data_criteria_references["GROUP_#{@children_criteria.first}"].nil? && @children_criteria.length == 1
-          @children_criteria[0] = "GROUP_#{@children_criteria.first}"
-        elsif @children_criteria.length == 1 && @children_criteria.first.present?
-          reference_criteria = @data_criteria_references[@children_criteria.first]
-          return if reference_criteria.nil?
-          duplicate_child_info(reference_criteria)
-          @children_criteria = reference_criteria.children_criteria
-        end
+        handle_do_not_group
         return
```
edeyoung: part of me feels like the `do_not_group` stuff should have some explanation here - I think it's because I'm in a method `extract_variable_grouper` and then it's like, "psych! I'm going to do something different!" but maybe it's not so different? I guess the point is, it's not clear to me what's happening here. This might be due to my general confusion about grouper data criteria.

Actually, looking at the `handle_do_not_group` code, it looks like it's only called here and it's relatively short. I think by moving that code into this function, it clarifies what is going, like what is meant by `do_not_group` and that this is still doing data extraction.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -392,388 +120,206 @@ def extract_as_grouper
...
-    # Patch SDC variables that are single data criteria by embedding the grouper
-    def patch_variable_data_criteria
-      return unless @variable && @is_source_data_criteria && !@derivation_operator
-      @derivation_operator = HQMF::DataCriteria::UNION
-      @definition = 'derived'
-      @status = nil
-      @children_criteria = ["GROUP_#{@id}"]
+    # Handles elments that only extract directly from the xml. Utilises the "BaseExtractions" class,
+    #  which holds a number of extractions that directly use the entry xml rather than implicitly
+    #  through the utilies module.
```
edeyoung: consider rewording to something like, 

"Handles elements that can be extracted directly from the xml."
At least this is how I interpreted this statement. Does this mean that other elements are extracted not from the XML or that extracting them from the XML is more complex and requires more code? If so, maybe put language like that in.

I'm not sure that the utilities vs baseextractions notes are necessary.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -392,388 +120,206 @@ def extract_as_grouper
...
-      # only consider non-variable specific occurrences
-      return unless @specific_occurrence && !@variable && !@id.start_with?("GROUP_")
-      reference = data_criteria_references[(strip_tokens @source_data_criteria)]
-      if reference && !reference.specific_occurrence && @specific_occurrence
-        # puts "Patching #{@id} from #{@source_data_criteria} to #{@id}."
-        @source_data_criteria = @id
+    # Extract the description, with some special handling if this is a variable; the MAT has added an encoded
+    # form of the variable name in the localVariableName field, if that's available use it; if not, fall back
+    # to the extension
+    def extract_description
```
edeyoung, line 169: should this method be in the `DataCriteriaBaseExtractions` class?

What's the difference between extraction methods in this class and the `DataCriteriaBaseExtractions` class?

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -392,388 +120,206 @@ def extract_as_grouper
...
-    private
-
-    def extract_negation
-      negation = attr_val('./*/@actionNegationInd')
-      @negation = (negation=='true')
-      if @negation
-        res =  @entry.at_xpath('./*/cda:outboundRelationship/*/cda:code[@code="410666004"]/../cda:value/@valueSet', HQMF2::Document::NAMESPACES)
-        @negation_code_list_id =  res.value if res
+    # Extract the code system from the xml taht the document should use
+    def inline_code_list
```
edeyoung: based on the description, this seems like it should be called something like `extract_code_system`.

Should this also be in the `DataCriteriaBaseExtractiosn` class?

lib/hqmf-parser/2.0/data_criteria.rb
elehner: Would it make it more clear to check for "do_not_group" in the document "handle_variable" function, and then have that call "handle_do_not_group" instead? (I'll also comment on what classifies a variable as a non-grouping variable)

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -392,388 +120,206 @@ def extract_as_grouper
...
-        Reference.new(ref).id
-      end.compact
-    end
-
-
-    def all_subset_operators
-      @entry.xpath('./*/cda:excerpt', HQMF2::Document::NAMESPACES).collect do |subset_operator|
-        SubsetOperator.new(subset_operator)
+    # Handle elements that are marked as variable groupers that should not be turned into a "holding element"
+    #  (defined as a data criteria that encapsulates the calculation material in another element, and it itself groups them together)
```
edeyoung: the explanation of a "holding element" is confusing to me.

lib/hqmf-parser/2.0/data_criteria.rb
elehner: The latter makes the most sense to me (the one long name).  The two actually go together, as the source data criteria information is parsed differently is there is a specific occurrence.

lib/hqmf-parser/2.0/data_criteria.rb
elehner, line 71: Should I add a comment to the main entry class (Document), or to every class?  The to_model function is somewhat generic, in that it generates the HQMF model for the class using the existing hqmf-model code.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -392,388 +120,206 @@ def extract_as_grouper
...
+      @subset_operators ||= child_ref.subset_operators
+      @variable ||= child_ref.variable
+      @value ||= child_ref.value
     end
 
-    def extract_subset_operators
-      all_subset_operators.select do |operator|
-        operator.type != 'UNION' && operator.type != 'XPRODUCT'
+    # Generate the models of the field values
+    def model_field_values
```
edeyoung: is this the same as extracting field values? If so, consider renaming for consistency/clarity. If not, leave as is.

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -392,388 +120,206 @@ def extract_as_grouper
...
-            # puts "\tSetting #{@source_data_criteria}-#{@source_data_criteria_root} to #{occurrenceIdentifier}"
-            @occurrences_map[@source_data_criteria] ||= occurrenceIdentifier
-          end
-          occurrence = @occurrences_map.try(:[], @source_data_criteria)
-          raise "Could not find occurrence mapping for #{@source_data_criteria}, #{@source_data_criteria_root}" unless occurrence
-          # puts "\tUsing #{occurrence} for #{@id}"
-          @specific_occurrence ||= occurrence
-        end
+    # Generate the title and description used when producing the model
+    def handle_title_and_description
```
edeyoung: same as above - is this the same as extracting the title/description? If so, consider renaming.

lib/hqmf-parser/2.0/data_criteria.rb
elehner: Is there a way it could be worded better?  They are functionally the same as the "GROUP_" elements present in the SimpleXML.  The way they are created is the extract_grouper method turns that data_criteria into a "GROUP_" criteria, and creates a new data criteria which references the "GROUP_" criteria as a child element.

lib/hqmf-parser/2.0/data_criteria.rb
elehner: The idea with the `DataCriteriaBaseExtractions` was to have a group of items that were extracted for use by this parsers data criteria class.  The inline_code_list is only used by the to_model, so it should probably be left there (the name change is a good point).

lib/hqmf-parser/2.0/data_criteria.rb
elehner, line 169: Since `extract_description` relies on `@variable`, I didn't feel it fell into the BaseExtractions domain.

lib/hqmf-parser/2.0/data_criteria.rb
elehner: It's different.  This uses the extracted field values and turns them into their proper HQMF model form, to be used by the Data Criteria's to_model function.

lib/hqmf-parser/2.0/data_criteria.rb
elehner: Not the same as extracting the title and description, it's also used only by the to_model function.  Thoughts on what name it should be? (I'm thinking to preface both with "to_model")

lib/hqmf-parser/2.0/data_criteria.rb
edeyoung, line 71: ah - this sounds like a ruby-ism that I'm not familiar with. if this is something that is used generally in ruby, has the same purpose across classes, and isn't an override or something, then I agree it doesn't do a whole lot of good to comment.

lib/hqmf-parser/2.0/data_criteria.rb
edeyoung: can a variable be a grouper variable and have `do_not_group` set to `true`? If so, I think it makes sense to leave it here. If not, then maybe it makes sense to move it out to the document level and rename it to `extract...`. 

If you keep it all in this class, I think the `handle_do_not_group` logic can be moved up into this function. It seems to mirror the other logic pretty closely, which helps you see the relationship between the two code paths.

lib/hqmf-parser/2.0/data_criteria.rb
elehner, line 71: It's commonly used in rails with ActiveRecord, but we define our own models (it's not necessarily a ruby-ism).  My point was that most of the classes we have (Document, Data Criteria, Populations, Preconditions, and every Type class) has a to_model method that performs the same/similar function, and I was wondering if it made more sense to comment the main Document to_model, or to comment every one of them.

lib/hqmf-parser/2.0/data_criteria.rb
edeyoung: I'm not clear on GROUP_ elements in SimpleXML, so the comparison is not helping me, unfortunately.

Some of the things I wasn't understanding in the first comment:
- what is meant by 'encapsulates the calculation material in another element'? Do you mean another data criteria? What do you mean by calculation material? Would this be the 'during', 'AND', etc. stuff? Maybe 'logic clauses' would make this more clear?
- what is meant by 'it itself group them together'? Does that mean that it contains a reference to whatever it is that contains the calculation material? I think I keep blocking on this idea of grouping and what that actually means.

lib/hqmf-parser/2.0/data_criteria.rb
elehner: I think I'll move that logic check to Document.  The issue was that some variables are not "grouper" variables, and the code initially marked all variables that way (though there is some additional handle for non-grouper variables that needs to take place).

lib/hqmf-parser/2.0/data_criteria.rb
edeyoung, line 71: My gut is all of them. If you're debugging, you might not get to the code through the document class. If that's the case, it's nice to have the appropriate context where you'll be debugging/working with the code.

lib/hqmf-parser/2.0/data_criteria.rb
edeyoung, line 169: ah, ok. I'll add a comment to the BaseExtractions file, but maybe add something to that about it not only being straightforward but self-contained? Or not. It just seemed funny to me at first pass that some extraction things were in a different class. What you said makes sense though.

lib/hqmf-parser/2.0/data_criteria.rb
edeyoung: or maybe `get_X_model`?
`to_model_X` would work.

I'll admit I'm confused by the `to_model` stuff vs what happens when you instantiate the object.

elehner: So the parser/instantiation is handling the parsing of the newest HQMF specification (it generates all the elements needed to handle it's own parsing).  The to_model is to turn that newly parsed data into a form we know how to handle for our calculations, and is relatively flexible enough to be used for version 1 and 2.

elehner: I'm using the syntax retrieve_X_for_model, as get can be a confusing prefix depending on the context.

lib/hqmf-parser/2.0/data_criteria_helpers/dc_post_processing.rb
```diff
@@ -0,0 +1,71 @@
+module HQMF2
+  # Processing on data criteria after the initial extractions have taken place
+  module DataCriteriaPostProcessing
+    # Handles settings values after (most) values have been setup
+    def post_processing
+      set_code_list_path_and_result_value
+
+      # prefix ids that start with numerical values, and strip tokens from others
+      @id = strip_tokens @id
+      @children_criteria.map! { |cc| strip_tokens cc }
```
edeyoung: consider adding parens here

lib/hqmf-parser/2.0/data_criteria_helpers/dc_post_processing.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      #### prefix!!!!!!!!!
```
edeyoung: what does this comment mean?

lib/hqmf-parser/2.0/data_criteria_helpers/dc_post_processing.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      @source_data_criteria = strip_tokens(@source_data_criteria) unless @source_data_criteria.nil?
+      @specific_occurrence_const = strip_tokens(@specific_occurrence_const) unless @specific_occurrence_const.nil?
+      set_intersection
+      handle_derived_specific_occurrences
+    end
+
+    # Extract the code_list_xpath from either the location related to the speicifc occurrence, or
+    #  from any of the template ids (if multiple exist)
+    def set_code_list_path_and_result_value
```
edeyoung: consider renaming to `extract_...`

Also, consider adding verbiage about the `@value` stuff going on to the comment. it was confusing to not have it in the description.

lib/hqmf-parser/2.0/data_criteria_helpers/dc_post_processing.rb
```diff
@@ -0,0 +1,71 @@
...
+    # Extract the value and code_list_xpath from the template mapping held in the ValueSetHelper class
+    def handle_mapping_template(mapping)
+      if mapping
+        @code_list_xpath = mapping[:valueset_path] if mapping[:valueset_path] && @entry.at_xpath(mapping[:valueset_path])
+        @value = DataCriteriaMethods.parse_value(@entry, mapping[:result_path]) if mapping[:result_path]
+      end
+    end
+
+    # Changes XPRODUCT data criteria that has an associated tempalte(s) to an INTERSETION criteria
+    def set_intersection
```
edeyoung: the function name is confusing based on what the method does. it looks like it only sets to intersection if it's a cross product. the function name looks like it will always set to intersection. consider renaming to something like, `change_xproduct_to_intersection`. 
Actually, this suggestion doesn't seem to capture that `Union` is used for everything else.

lib/hqmf-parser/2.0/data_criteria_helpers/dc_specific_occurrences_and_source_data_criteria_extract.rb
```diff
@@ -0,0 +1,108 @@
...
+      else
+        occurrence_lvn_regex = 'Occurrence[A-Z]of'
+        occurrence_id_regex = 'Occurrence[A-Z]_'
+        occ_index = 10
+
+        if stripped_id.match(/^#{occurrence_id_regex}#{stripped_sdc}/)
+          return stripped_id[occ_index]
+        elsif stripped_lvn.match(/^#{occurrence_lvn_regex}#{stripped_sdc}/)
+          return stripped_lvn[occ_index]
+        end
```
edeyoung: this code block looks the same as `handle_occurrence_var`. consider using the method.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -1,202 +1,28 @@
...
-          @stratifications << stratified_population
-        end
-      end
-
-      # Push in the stratification populations after the unstratified populations
-      @populations.push *@stratifications
+      pop_helper = HQMF2::DocumentPopulationHelper.new(@entry, @doc, self, @id_generator, @reference_ids)
+      @populations, @population_criteria = pop_helper.extract_populations_and_criteria
 
       @reference_ids.uniq
```
edeyoung: consider adding a comment explaining how `@reference_ids` are used. This might not be the right spot, but it's the firs time I've seen it.

edeyoung: actually, consider moving this line into `covered_criteria?`. `covered_criteria?` is explained and it's basically why `@reference_ids` are used.

lib/hqmf-parser/2.0/data_criteria_helpers/dc_post_processing.rb
elehner: Are you saying `@children_criteria.map!() { |cc| strip_tokens cc }`?

lib/hqmf-parser/2.0/data_criteria_helpers/dc_post_processing.rb
elehner: Possible leftover from previous code. Removed.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -239,7 +65,7 @@ def data_criteria(id)
       find(@data_criteria, :id, id)
     end
 
-    #needed so data criteria can be added to a document from other objects
+    # needed so data criteria can be added to a document from other objects
```
edeyoung, line 71: for `find_criteria_by_lvn` (for some reason I can't comment down there because no changes were made there), add a comment. For me, at least, `lvn` doesn't automatically translate to local variable name.

Also, `add_data_criteria` and `find_criteria_by_lvn` don't have the formal comments that the rest of the methods have. They're pretty self-explanatory, so this would just be a consistency thing more than a clarity thing.

lib/hqmf-parser/2.0/data_criteria_helpers/dc_specific_occurrences_and_source_data_criteria_extract.rb
elehner: Done

lib/hqmf-parser/2.0/data_criteria_helpers/dc_post_processing.rb
edeyoung: `@children_criteria.map!  { |cc| strip_tokens(cc) }`

For both this line and the line above.

lib/hqmf-parser/2.0/data_criteria_helpers/dc_post_processing.rb
elehner: But it does work for the function name.  I'll add the "Union is used for everything else" comment to the function comment.

edeyoung: to me, `set_intersection` would mean that whatever I called this on, `Intersect` would be used in the description and as the derivation_operator. Since that's not the case, I found the function name confusing.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -251,188 +77,85 @@ def find_criteria_by_lvn(lvn)
...
 
-    # If a precondition references a population, remove it
-    def remove_population_preconditions(doc)
-      #population sections
-      pop_ids = doc.xpath("//cda:populationCriteriaSection/cda:component[@typeCode='COMP']/*/cda:id",NAMESPACES)
-      #find the population entries and get their ids
-      pop_ids.each do |p_id|
-        doc.xpath("//cda:precondition[./cda:criteriaReference/cda:id[@extension='#{p_id["extension"]}' and @root='#{p_id["root"]}']]",NAMESPACES).remove
-      end
+    def find(collection, attribute, value)
```
edeyoung, line 101: consider adding a comment - it's not clear what I'm finding in what.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -251,188 +77,85 @@ def find_criteria_by_lvn(lvn)
...
-          end
-          ids_by_hqmf_id["#{criteria.hqmf_id}-#{population['stratification']}"] = criteria.id
-        end
-
-
-        @population_criteria << criteria
-        population[criteria_id] = criteria.id
-      else
-        population[criteria_id] = identical.first.id
+    def extract_preconditions(precondition, list)
```
edeyoung: consider adding a comment explaining why you only extract a precondition `if prcn.reference`.

lib/hqmf-parser/2.0/data_criteria_helpers/dc_post_processing.rb
elehner: Modified for all cases throughout the classes.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -251,188 +77,85 @@ def find_criteria_by_lvn(lvn)
...
-      base_checks &&= data_criteria.subset_operators.empty? # Ensure it doesn' it have a subset operator
-
-      same_value = data_criteria.value.nil? || data_criteria.value.try(:to_model).try(:to_json) == check_criteria.value.try(:to_model).try(:to_json)
-      same_temporal_references = data_criteria.temporal_references.nil? || data_criteria.temporal_references.empty?
-      same_field_values = data_criteria.field_values.nil? || data_criteria.field_values.empty? || data_criteria.field_values.try(:to_json) == check_criteria.field_values.try(:to_json)
-      same_negation_values = data_criteria.negation_code_list_id.nil? || data_criteria.negation_code_list_id == check_criteria.negation_code_list_id
-
-      base_checks && same_value && same_temporal_references && same_field_values && same_negation_values
-    end
+    # handles setup of the base values of the document, defined here as ones that are either obtained from the xml directly or with limited parsing
```
edeyoung: several comments in this file are very long-lined. consider breaking across multiple lines for readability.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -441,63 +164,107 @@ def read_attribute(attribute)
...
-          value_obj = value.present? ? HQMF::GenericValueContainer.new(type, value) : HQMF::AnyValue.new(type)
+      value_obj = handle_attribute_value(attribute, value) if attribute.at_xpath('./cda:value', NAMESPACES)
+
+      # Handle the cms_id
+      @cms_id = "CMS#{value}v#{@hqmf_version_number}" if name.include? 'eMeasure Identifier'
+
+      HQMF::Attribute.new(id, code, value, nil, name, id_obj, code_obj, value_obj)
+    end
+
+    # Extracts the code used by a specific attribute
```
edeyoung: this may seem a bit silly, but because "specific" has special meaning with this measure stuff, consider changing to "particular" or something similar.

edeyoung: when is `handle` used vs `extract`? If it's the same all around, consider changing all to `extract`. If there's a difference, consider changing the comments to avoid confusion.

lib/hqmf-parser/2.0/document.rb
```diff
@@ -441,63 +164,107 @@ def read_attribute(attribute)
...
+
+      extracted_criteria.each do |entry|
+        criteria = DataCriteria.new(entry, @data_criteria_references, @occurrences_map)
+        handle_data_criteria(criteria, collapsed_source_data_criteria)
+        @data_criteria << criteria
+      end
+    end
 
+    def handle_data_criteria(criteria, collapsed_source_data_criteria)
+      # Sometimes there are multiple criteria with the same ID, even though they're different; in the HQMF
```
edeyoung, line 247: As an aside, is this something we should mention to the MAT team? It doesn't seem like this should be the case.

lib/hqmf-parser/2.0/document.rb
elehner, line 71: I can change the parameter name, but the function is actually in use by a few test cases, so it's pretty difficult to change.  LVN is also a common acronym used throughout the code, it may not be worth it to change.

edeyoung, line 71: sorry, i meant just add a comment at the top of the function like, 

```
# Gets the data criteria by the local variable name.
```

elehner, line 71: Ah, okay, misread.

lib/hqmf-parser/2.0/document.rb
elehner: Just removed this function, as the function that originally called it no longer exists (wasn't being used).

lib/hqmf-parser/2.0/document_helpers/doc_population_helper.rb
```diff
@@ -0,0 +1,165 @@
...
+        handle_stratifications(population_def, number_of_populations, population, id_def, population_index)
+      end
+
+      # Push in the stratification populations after the unstratified populations
+      @populations.concat(@stratifications)
+      [@populations, @population_criteria]
+    end
+
+    # Extracts the measure observations, will return true if one is created
+    def extract_observation
```
edeyoung: consider revising the name to `extract_observations`.

consider changing the comment to `.... will return true if one exists.` As written, I thought it would be creating a measure observation if one didn't already exist.

lib/hqmf-parser/2.0/document.rb
elehner: I tried to use handle in two cases: 1) the code modifies instance variables, but doesn't directly touch the xml, and 2) the code touches the xml, but does not have that information placed into an instance variable. Essentially, extract covers the case where elements are being taken from the xml to be stored in an instance variable.

lib/hqmf-parser/2.0/document_helpers/doc_utilities.rb
```diff
@@ -0,0 +1,78 @@
...
+      return unless grouper_data_criteria
+      @data_criteria_references[data_criteria.id] = data_criteria
+      @data_criteria_references[grouper_data_criteria.id] = grouper_data_criteria
+      @data_criteria << grouper_data_criteria
+      @source_data_criteria << SourceDataCriteriaHelper.strip_non_sc_elements(grouper_data_criteria)
+    end
+
+    # Update the data criteria to handle variables properly
+    def update_data_criteria(data_criteria, source_data_criteria)
+      # step through each criteria and look for groupers (type derived) with one child
```
edeyoung: this comment leaves me wanting more. consider adding something like, "then add the matching source_criteria children to the criteria and update .... <and maybe something about how it's related to variables? is the child the variable?>"

lib/hqmf-parser/2.0/document_helpers/doc_utilities.rb
```diff
@@ -0,0 +1,78 @@
...
+      to_reject && !@data_criteria.detect do |dc2|
+        similar_criteria = true
+        similar_criteria &&= dc != dc2 # Don't check against itself
+        similar_criteria &&= dc.code_list_id == dc2.code_list_id # Ensure code list ids are the same
+        similar_criteria && detect_criteria_covered_by_another(dc, dc2)
+      end.nil? # don't reject unless there is a similar element
+    end
+
+    # Check if one data criteria contains the others information by checking that one has everything the other has (or more)
+    def detect_criteria_covered_by_another(data_criteria, check_criteria)
```
edeyoung: the name of this method doesn't seem differentiated enough from `covered_criteria?`. maybe `criteria_covered_by_criteria?`. To me, this indicates more that I'm talking about a specific given criteria than a general, any criteria.

lib/hqmf-parser/2.0/document_helpers/doc_utilities.rb
```diff
@@ -0,0 +1,78 @@
...
+      # Ensure it doesn't contain basic elements that should not be removed
+      base_checks &&= !data_criteria.variable # Ensure it's not a variable
+      base_checks &&= data_criteria.derivation_operator.nil? # Ensure it doesn't it have a derivation operator
+      base_checks &&= data_criteria.subset_operators.empty? # Ensure it doesn't it have a subset operator
+      base_checks &&= data_criteria.temporal_references.nil? || data_criteria.temporal_references.empty? # Ensure it doesn't have Temporal References
+
+      base_checks && complex_coverage(data_criteria, check_criteria)
+    end
+
+    # Check elements that need to not only exist, but need to be non-equal to other, potentially matchin, data criteria.
```
edeyoung: this comment is confusing to me.

Maybe "check elements that must exist and must not equal elements of other potentially  matching data criteria"

lib/hqmf-parser/2.0/document.rb
elehner, line 247: Quick test found this issue exists only for CMS169 (but is alleviated when using the most recent criteria with a particular id).

lib/hqmf-parser/2.0/document.rb
edeyoung: that makes sense. In that case, I'd change the comments to not use "extract". It makes it seem like it should serve the same purpose as the `extract` functions.

lib/hqmf-parser/2.0/document_helpers/doc_utilities.rb
elehner: I can change it, but it looks like the internal structure of that function that handles the update is never reached (at least not over our test set), which sets this up for potential removal. Thoughts?

edeyoung: I don't feel like I'm familiar enough with this code to say. If it's actually never reached, then I think it makes sense to remove it. Are the test sets broad enough to give us this assurance? I'd leave this to Pete.

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -1,177 +1,206 @@
 module HQMF2
-
+  # Generates field values based on understanding of the HQMF 2.1 spec
   class FieldValueHelper
-    def self.parse_field_values(entry,negated=false)
-      return  if entry.nil?
-      criteria = entry.at_xpath("./cda:actCriteria | ./cda:observationCriteria | ./cda:encounterCriteria | ./cda:procedureCriteria | ./cda:supplyCriteria | ./cda:substanceAdministrationCriteria | ./cda:grouperCriteria")
+    def self.parse_field_values(entry, _negated = false)
```
edeyoung: can we remove `_negated`?

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -1,177 +1,206 @@
...
-        parse_substance_administration_fields(criteria,fields)
+        parse_substance_administration_fields(criteria, fields)
       when 'grouperCriteria'
-        parse_grouper_fields(criteria,fields)
+        parse_grouper_fields(criteria, fields)
       end
-      fields
     end
 
+    # The next groups of "parse_" handles extracting values for different types of criteria
```
edeyoung: Consider adding a space. My personal preference for these types of comments (comments referring to multiple functions) is to leave a space between them and the next function. Without the space, it seems tightly tied to the function it's above.

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -1,177 +1,206 @@
...
         any = any_flavor(item) || any_flavor(element)
         fields[field] = AnyValue.new if any
         fields[field] = Coded.new(item) if item && !any
       end
     end
 
-    def self.parse_cd(element,field,fields)
-      any = any_flavor(element)
-      fields[field] = AnyValue.new if any
+    # These "parse_" methods also first check if the value is of the "ANY" type, and only parse if it isn't
```
edeyoung: consider adding a space. see comment above.

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -1,177 +1,206 @@
...
+    def self.parse_ivl_int(element, field, fields)
+      any = check_if_any(element, field, fields)
       fields[field] = Range.new(element) if element && !any
     end
 
-    def self.parse_cs(element,field,fields)
-      any = any_flavor(element)
-      fields[field] = AnyValue.new if any
+    def self.parse_cs(element, field, fields)
+      check_if_any(element, field, fields)
```
edeyoung: this function behaves differently than the others. I'd add a comment explaining that.

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -1,177 +1,206 @@
...
-      )
+    # handle any value
+    def self.parse_any(element, field, fields)
+      any = check_if_any(element, field, fields)
+      fields[field] = DateCriteria.parse_value(element) if element && !any
     end
 
-    def self.parse_value(element,field,fields)
+    # Use when checking if the element is a "ANY" type, sets the field key if it is
+    def self.check_if_any(element, field, fields)
```
edeyoung: consider renaming this function. When I was reading the functions above, I thought it was just doing a check to see if the element was `any`. I see here that it's actually setting the field key. this is not clear from how the functions above flow. maybe `check_and_set_if_any`.

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -1,177 +1,206 @@
...
+        'REMOVAL_DATETIME'
+      elsif template_ids.include?('2.16.840.1.113883.10.20.28.3.110') || template_ids.include?('2.16.840.1.113883.10.20.28.3.116')
+        'ABATEMENT_DATETIME'
+      else
+        'DISCHARGE_DATETIME'
+      end
+    end
 
-    def self.parse_act_criteria_fields(entry,fields)
+    # The "parse_"s after this point handle extraction of data criteria based on field names
```
edeyoung, line 174: same comment as above re comments that refer to multiple functions

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -1,177 +1,206 @@
...
+    def self.parse_act_criteria_fields(_entry, _fields)
```
edeyoung, line 176: can this be removed?

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -1,177 +1,206 @@
...
+               'REFERENCE_RANGE_HIGH', fields)
+      parse_pq(entry.at_xpath("./cda:outboundRelationship[@typeCode='REFV']/cda:observationCriteria/cda:value/cda:low", HQMF2::Document::NAMESPACES),
+               'REFERENCE_RANGE_LOW', fields)
     end
 
-    def self.parse_encounter_fields(entry,fields)
-      #Added a check for Principal Diagnosis and Diagnosis. QDM 4.2 Update
-      principal = entry.at_xpath("./cda:outboundRelationship[@typeCode='REFR']/cda:actCriteria/cda:code[@code='52534-5']", HQMF2::Document::NAMESPACES)
+    # Ignoring line limits here as it would be hard to create the deep xPaths with these limits.
+    # rubocop:disable Metrics/LineLength
```
edeyoung, line 206: i have a few personal style thoughts here.
1. my gut is that if you'll be ignoring line lengths, to do it for the whole file
2. I was surprised that line lengths were not already being ignored given how long the lines were, so I looked at the rubocop settings. It looks like the line lengths are set to 150. My personal feeling is that is too big. If I'm just on my laptop with no extenral monitor and one window open in atom, the max line lengths I can handle without scrolling are ~120. If I'm reviewing in github, the max line lengths it shows are also ~120. My personal preference would be ~100, but I think ~120 makes sense given the above.

lib/hqmf-parser/2.0/document_helpers/doc_utilities.rb
pkmitre: We definitely want to remove dead code. Our test for dead code needs to include the 2015 measures (which I know is already used) as well as a selection of more recent MAT output.

lib/hqmf-parser/2.0/field_value_helper.rb
```diff
@@ -1,177 +1,206 @@
...
-      parse_cd(entry.at_xpath('./cda:dischargeDispositionCode', HQMF2::Document::NAMESPACES),'DISCHARGE_STATUS',fields)
+      parse_pq(entry.at_xpath('./cda:lengthOfStayQuantity', HQMF2::Document::NAMESPACES), 'LENGTH_OF_STAY', fields)
+      parse_cd(entry.at_xpath('./cda:dischargeDispositionCode', HQMF2::Document::NAMESPACES), 'DISCHARGE_STATUS', fields)
 
-      loc =  entry.at_xpath("./cda:participation[@typeCode='LOC']/cda:role[@classCode='SDLOC']", HQMF2::Document::NAMESPACES)
+      handle_loc(entry, fields)
+    end
+    # rubocop:endable Metrics/LineLength
+
+    def self.handle_loc(entry, fields)
```
edeyoung: consider moving this out of the group of `parse_` functions listed here and included in the overall comment above. after moving it, consider adding a comment to help people mind-shift away from the `parse_` functions.

lib/hqmf-parser/2.0/field_value_helper.rb
elehner, line 176: It potentially can.  It's essentially a placeholder (used in the `handle_fields_per_criteria` function), and I left it there as such.  Should we remove it?

lib/hqmf-parser/2.0/population_criteria.rb
```diff
@@ -14,75 +13,90 @@ def initialize(entry, doc, id_generator)
...
-          @title = attr_val('../cda:code/cda:displayName/@value')
-          @aggregator = attr_val('./cda:measureObservationDefinition/cda:methodCode/cda:item/@code')
-          # MEAN is handled in current code. Changed since it should have the same effect
-          @aggregator = 'MEAN' if @aggregator == 'AVERAGE'
-      end
-      if(!@hqmf_id) # The id extension is not required, if it's not provided use the code
-        @hqmf_id = @type
-      end
+      # If there are no measure observations, or there is a title, then there are no aggregations to extract
+      return unless !@title && obs_test.to_s == 'OBS'
```
edeyoung, line 51: is there guidance re `return unless`? I feel like this would read better as:

```
      handle_preconditions(id_generator)
      obs_test = attr_val('./cda:measureObservationDefinition/@classCode')
      # If there are no measure observations, or there is a title, then there are no aggregations to extract
      if !@title && obs_test.to_s == 'OBS'
        @title = attr_val('../cda:code/cda:displayName/@value')
        @aggregator = attr_val('./cda:measureObservationDefinition/cda:methodCode/cda:item/@code')
      end
```

lib/hqmf-parser/2.0/population_criteria.rb
```diff
@@ -14,75 +13,90 @@ def initialize(entry, doc, id_generator)
...
-        dc = handle_observation_critiera
-        @preconditions = [Precondition.new(id_generator.next_id,nil, nil, false, HQMF2::Reference.new(dc.id))]
-      end
     end
 
-
+    # extracts out any measure observation definitons, creating from them the proper criteria to generate a precondition
     def handle_observation_critiera
+      exp = @entry.at_xpath('./cda:measureObservationDefinition/cda:value/cda:expression/@value', HQMF2::Document::NAMESPACES)
+      fail 'No Expression ' if exp.nil?
```
edeyoung: should this include more information about what has no expression, or does a fail in ruby include the appropriate context automatically (like the stack trace)?

lib/hqmf-parser/2.0/population_criteria.rb
```diff
@@ -14,75 +13,90 @@ def initialize(entry, doc, id_generator)
...
+        # If there are neither one or 2 parts, the code should fail
+        fail "No defined extraction method to handle #{parts.length} parts"
       end
-
-      @doc.add_data_criteria(dc)
-      dc
-
     end
 
     def create_human_readable_id(id)
```
edeyoung, line 107: in `create_human_readable_id`, it doesn't change anything, just sets a variable. Should it be renamed?

in `conjunction?`, the description doesn't match the logic and the logic always returns true. consider rewording the comment or getting rid of the function.
(`conjunction?` is a few lines below and github won't let me comment there)

lib/hqmf-parser/2.0/population_criteria.rb
elehner, line 51: The idea is to use a "guard statement" (http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/GuardClause) instead of using another conditional at the end.  We can change it, but this has been used in a number of places in the code already, so it may make more sense to leave it.

lib/hqmf-parser/2.0/population_criteria.rb
```diff
@@ -99,20 +113,20 @@ def conjunction?
     # @return [String] conjunction code
```
edeyoung, line 112: looks like the comment doesn't match the logic. only `ALL_TRUE` and `AT_LEAST_ONE_TRUE` are returned. consider rewriting the comment.
(relevant comment is also in an area github doesn't let me comment on)

lib/hqmf-parser/2.0/field_value_helper.rb
elehner, line 206: On ignoring for the whole file, I think it's better if we scope it to a particular piece of code.  If we ignore for the entire file, it'll be confusing as to why we ignore it at all, but with the single function ignored it becomes very visible as to what necessitates the ignore statement (the long xPath string).

The 150 line length limit was to better match the styling's that cypress's cypress3 branch uses. We can set it to 100, but that will incur a larger number of rewrites throughout this code base.

lib/hqmf-parser/2.0/source_data_criteria_helper.rb
```diff
@@ -1,26 +1,30 @@
...
       sha256 << "#{criteria.variable}:"
-      sha256 << (criteria.children_criteria.nil? ? "<nil>:" : "#{criteria.children_criteria.sort.join(',')}:")
+      sha256 << (criteria.children_criteria.nil? ? '<nil>:' : "#{criteria.children_criteria.sort.join(',')}:")
 
       Digest::SHA256.hexdigest sha256
-  	end
+    end
 
+    # Rejects any derived elements as they should never be used as source.
     def self.should_reject(dc)
```
edeyoung: should this be named something like `.reject?` or `.should_reject?`?

lib/hqmf-parser/2.0/types.rb
```diff
@@ -1,5 +1,4 @@
 module HQMF2
-
   # TODO: Clean up debug print statements!
```
edeyoung: have these been cleaned up? if so, consider removing comment.

lib/hqmf-parser/2.0/population_criteria.rb
elehner, line 107: I believe the name is supposed to reinforce that this function is used to update the id to something human readable, since the code cannot enforce that itself.

lib/hqmf-parser/2.0/field_value_helper.rb
edeyoung, line 176: @pkmitre, thoughts?

lib/hqmf-parser/2.0/field_value_helper.rb
edeyoung, line 206: 1. that makes sense
2. I'd push for 120. I know it would be tedious to change and I'm happy to implement it. I feel it would make the code significantly more readable (at least for me). @pkmitre, thoughts?

lib/hqmf-parser/2.0/population_criteria.rb
edeyoung, line 51: If the style guide says to use it, then let's keep it.
(I'm finding I frequently disagree with ruby style guides... oh well... I'll get used to it eventually :) )

lib/hqmf-parser/2.0/population_criteria.rb
elehner, line 112: Corrected

lib/hqmf-parser/2.0/field_value_helper.rb
pkmitre, line 206: I suspect that in this code a lot of the long line lengths are due to xpath statements, which would be less readable if broken up.  I am OK with shortening up individual long lines for readability, but I don't think we need to do it holistically for this entire code base.  We should reconsider for new code.

PR #339, opened 2015-12-27
Added ticket_granting_ticket param to VSAC api

holmesie:
When initializing a VSApi or VSApiV2 object, a ticket granting ticket can now optionally be specified, which will be used to retrieve service tickets for all VSAC API calls (as opposed to requesting a new tgt every single time an API call is made).

### DIFF
--- a/lib/health-data-standards/util/vs_api.rb
+++ b/lib/health-data-standards/util/vs_api.rb
@@ -8,11 +8,12 @@ module HealthDataStandards
 		class VSApi
 			attr_accessor :api_url, :ticket_url, :username, :password
 
-			def initialize(ticket_url, api_url, username, password)
+			def initialize(ticket_url, api_url, username, password, tgt=nil)
 				@api_url = api_url
 				@ticket_url = ticket_url
 				@username = username
 				@password = password
+        @proxy_ticket = tgt
 			end
 
       def get_valueset(oid, effective_date=nil, include_draft=false, &block)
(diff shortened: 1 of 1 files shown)


lib/health-data-standards/util/vs_api.rb
```diff
@@ -8,11 +8,12 @@ class VSNotFoundError < StandardError
 		class VSApi
 			attr_accessor :api_url, :ticket_url, :username, :password
 
-			def initialize(ticket_url, api_url, username, password)
+			def initialize(ticket_url, api_url, username, password, tgt=nil)
```
pkmitre: How unwieldy would it be to expand tgt to ticket_granting_ticket? I find tgt a bit opaque.

lib/health-data-standards/util/vs_api.rb
```diff
@@ -8,11 +8,12 @@ class VSNotFoundError < StandardError
...
 				@api_url = api_url
 				@ticket_url = ticket_url
 				@username = username
 				@password = password
+        @proxy_ticket = tgt
```
pkmitre: What's @proxy_ticket being used for? I don't see it referenced in this file.

pkmitre: Never mind, I see it now in the proxy_ticket method, not sure how I missed that.

lib/health-data-standards/util/vs_api.rb
holmesie: It might be somewhat hard on line lengths, but I do now think it might make things much more clear. This is also something that I think should be changed in the corresponding bonnie_bundler pull request.

PR #342, opened 2016-01-06
Bonnie 104932540 update value sets final

alexanderelliott121:


lib/health-data-standards/models/svs/concept.rb
```diff
@@ -12,6 +12,13 @@ class Concept
       field :code_system, type: String
       scope :by_code_system_name,  ->(cs){where(code_system_name: cs)}
       scope :by_code_system,  ->(cs){where(code_system: cs)}
+      
+      def ==(comp)
+        if (self.code != comp.code)
+          return false
+        end
+        return true
+      end
```
pkmitre, line 18: Are we still using this with the new approach? If so, it could be simplified as

```
def ==(comp)
  return self.code == comp.code
end
```

(And if not I guess we can remove it)

lib/health-data-standards/models/svs/value_set.rb
```diff
@@ -5,7 +5,7 @@ class ValueSet
       field :oid, type: String
       field :display_name, type: String
       field :version, type: String
-      field :user_id, type: String # Eventually we need to delete this from bundles when exporting
+      field :bonnie_hash, type: String
```
pkmitre: Is there a more descriptive name we can use for this? version_hash? bonnie_version_hash? uniqueness_hash? unique_version_hash?

lib/health-data-standards/models/svs/value_set.rb
```diff
@@ -35,14 +39,24 @@ def code_set_map
         codes
       end
 
+      def self.gen_bonnie_hash(value_set)
```
pkmitre: I'd like to see "gen" expanded out to "generate". For an extra 5 characters we make sure that no one looks at this and wonders... general? generic? gentle? :->

lib/health-data-standards/models/svs/value_set.rb
```diff
@@ -35,14 +39,24 @@ def code_set_map
...
+        hash_values.unshift(value_set.oid)
+        bonnie_hash = Digest::MD5.hexdigest(hash_values.join('|'))
+        bonnie_hash
+      end
+
+
       def self.load_from_xml(doc)
         doc.root.add_namespace_definition("vs","urn:ihe:iti:svs:2008")
-        vs_element = doc.at_xpath("/vs:RetrieveValueSetResponse/vs:ValueSet|/vs:RetrieveMultipleValueSetsResponse/vs:DescribedValueSet")
+        vs_element = doc.at_xpath("/vs:RetrieveValueSetResponse/vs:ValueSet")
```
pkmitre: It looks like you're reverting a change made by @hossenlopp in an earlier pull request?

lib/health-data-standards/util/vs_api.rb
```diff
@@ -47,6 +47,7 @@ def get_ticket
 		end
 
     class VSApiV2 < VSApi
+
```
pkmitre: Ideally we'd not include files without substantive changes

pkmitre:
Given that this pull request turned out quite concise, I'd prefer to have all the changes squashed into a single commit.

lib/health-data-standards/models/svs/concept.rb
alexanderelliott121, line 18: I just realized that his should actually be comparing both code and code_system

lib/health-data-standards/util/vs_api.rb
alexanderelliott121: Ideally, yes.  There was a point where there were substantial changes, then they ceased to be necessary.  I wasn't able to find a way to remove the file from changes.

lib/health-data-standards/models/svs/value_set.rb
alexanderelliott121: Confusion on my part, will revert it back

lib/health-data-standards/models/svs/concept.rb
```diff
@@ -12,6 +12,16 @@ class Concept
...
+      
+      def ==(comp)
+        if (self.code != comp.code)
+          return false
+        end
+        if (self.code_system_name != comp.code_system_name)
+          return false
+        end
+        return true
+      end
```
pkmitre, line 18: I think this would be simpler and clearer as

```
def ==(comp)
  self.code == comp.code && self.code_system_name == comp.code_system_name
end
```

lib/health-data-standards/models/svs/value_set.rb
```diff
@@ -5,7 +5,7 @@ class ValueSet
       field :oid, type: String
       field :display_name, type: String
       field :version, type: String
-      field :user_id, type: String # Eventually we need to delete this from bundles when exporting
+      field :bonnie_version_hash, type: String #incoproates oid, version and concepts
```
pkmitre: Standard comment style for ruby includes a space between the # and the text

lib/health-data-standards/models/svs/value_set.rb
```diff
@@ -35,6 +41,17 @@ def code_set_map
         codes
       end
 
+      #Bonnie hash is based on oid, version and sorted concepts
+      def self.generate_bonnie_hash(value_set)
```
pkmitre, line 47: Why a class method rather than an instance method?

alexanderelliott121, line 47: Largely a matter of personal coding style.  This implementation allows a bonnie_version_hash to be generated without setting the bonnie_version_hash of the value_set it is derived from, which could theoretically be useful at some point in the future.

pkmitre, line 47: That goal could still be accomplished with an instance method, it would just return the value without setting anything.

lib/health-data-standards/models/svs/value_set.rb
```diff
@@ -19,6 +19,13 @@ class ValueSet
       index "concepts.display_name" => 1
       index "bundle_id" => 1
       scope :by_oid, ->(oid){where(:oid => oid)}
+      scope :by_bonnie_version_hash, ->(bonnie_version_hash){where(:bonnie_version_hash => bonnie_version_hash)}
+
+      before_save do |document|
+        if (document.bonnie_version_hash.nil?)
+          document.bonnie_version_hash = HealthDataStandards::SVS::ValueSet.generate_bonnie_hash(document)
+        end
+      end
```
pkmitre, line 28: Something like this is more in line with how we've built save hooks elsewhere in the code:

```
before_save :generate_bonnie_version_hash
def generate_bonnie_version_hash
  return self.bonnie_version_hash if self.bonnie_version_hash
  # Bonnie hash is based on oid, version and sorted concepts
  hash_values = value_set.concepts.map { |c| [c.code_system_name, c.code] }.sort.flatten
  hash_values.unshift(value_set.version)
  hash_values.unshift(value_set.oid)
  self.bonnie_version_hash = Digest::MD5.hexdigest(hash_values.join('|'))
end
```

There's a question about whether we should update the hash on save; updating the hash is in some ways more robust if the value set changes... but one could argue that a value set, once saved, should never change.

PR #343, opened 2016-01-07
Fixed include_draft option in VSApiV2.

hossenlopp:
- Added a default profile parameter when using the include_draft option.
  - The profile parameter is needed by VSAC V2 when using includeDraft.
- Added an option to get_valuests so a profile can be explicity provided.
- Fixed the class name of the VSApi tests to not conflict with another class of tests.
- Added unit tests for the cases dealing with include_draft and profile.

### DIFF
--- a/lib/health-data-standards/util/vs_api.rb
+++ b/lib/health-data-standards/util/vs_api.rb
@@ -66,12 +66,12 @@ module HealthDataStandards
       def get_valueset(oid, options = {}, &block)
         version = options.fetch(:version, nil)
         include_draft = options.fetch(:include_draft, false)
-        profile = options.fetch(:profile, nil)
+        profile = options.fetch(:profile, DEFAULT_PROFILE)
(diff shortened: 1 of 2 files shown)


lib/health-data-standards/util/vs_api.rb
```diff
@@ -59,10 +66,12 @@ def initialize(ticket_url, api_url, username, password, ticket_granting_ticket =
       def get_valueset(oid, options = {}, &block)
         version = options.fetch(:version, nil)
         include_draft = options.fetch(:include_draft, false)
+        profile = options.fetch(:profile, nil)
```
pkmitre: The second argument to fetch is the default to provide if the key isn't present, so this could be simplified by changing this to

```
profile = options.fetch(:profile, DEFAULT_PROFILE)
```

and then tweaking the setting of :profile in params below to

```
params[:profile] = profile if include_draft
```

PR #346, opened 2016-01-11
Small fixes to C32 and CCDA templates to address incorrect code_display rendering

ewhitley:
While I was eyeballing the template output I noticed these were not rendering "codes" as anticipated.   Templates with valid preferred code sets were rendering using the translation codes instead.

EX: 
## Should have rendered:

``` html
<code code="225171007" codeSystem="2.16.840.1.113883.6.96"><originalText>BMI</originalText></code>
```
## But was instead rendering:

``` html
<code nullFlavor="UNK"><originalText>BMI</originalText><translation code="225171007" codeSystem="2.16.840.1.113883.6.96"/>
</code>
```

Tracked this back to two separate issues across the templates.
(1) Medical Equipment: Symbol ":preferred_code_sets" sent instead of string for "'preferred_code_sets'"
(2) Vitals: "SNOMED" (not a valid code-system alias) was being sent instead of "SNOMED-CT"

All tests passed and I don't think this should break anything else since this is exclusively a template rendering fix.


pkmitre:
Merged via e02221a93d01fcb1b261ec5711860b9acd8d8027

PR #355, opened 2016-02-15
Updates to templates for R3 and expanded bundle

dczulada:


templates/cat1/_2.16.840.1.113883.10.20.24.3.9.cat1.erb
```diff
@@ -6,13 +6,19 @@
...
 
-    <effectiveTime>
+    <effectiveTime xsi:type="IVL_TS">
       <low <%= value_or_null_flavor(entry.start_time) %>/>
       <high <%= value_or_null_flavor(entry.removal_time || entry.end_time) %>/>
     </effectiveTime>
-
+    <author>
+      <templateId root="2.16.840.1.113883.10.20.22.4.119"/>
+      <time <%= value_or_null_flavor(entry.as_point_in_time) %>/>
```
okeefm, line 17: Should `author` have a `low` and a `high` time? According to the IG, 

> author/time/low represents the time the order is placed and author/time/high represents the time the order placement is completed.

PR #356, opened 2016-02-18
Mat format change

holmesie:


pkmitre:
I'm wondering if we should continue including the old xpath as well as an alternate (I guess using xpath or syntax) so that existing broken measures will continue loading... let me do a bit of digging.

pkmitre:
Turns out there aren't any existing measures that use these templates.

PR #361, opened 2016-02-23
Added address and name info to provider template

okeefm:


templates/cat1/_providers.cat1.erb
```diff
@@ -39,19 +39,35 @@
...
           <representedOrganization>
             <% pp.provider.cda_identifiers.each do |cda_id| -%>
             <% if cda_id.root.eql?('2.16.840.1.113883.4.2') -%>
-            <id root="2.16.840.1.113883.4.2" extension="<%= cda_id.extension %>" /> 
-            <% end -%>  
+            <id root="2.16.840.1.113883.4.2" extension="<%= cda_id.extension %>" />
             <% end -%>
+            <% end -%>
+            <%if pp.provider.organization %>
+              <name>pp.provider.organization.name</name>
```
dehall: should be `<%= %>` tags around pp.provider.organization.name

okeefm: Fixed; thanks for the catch.

PR #363, opened 2016-02-24
Updates for QRDA Cat I R3

dczulada:


lib/health-data-standards/import/cda/section_importer.rb
```diff
@@ -189,6 +189,25 @@ def extract_negation(parent_element, entry)
           end
         end
 
+        def extract_reason(parent_element, entry)
+          negation_indicator = parent_element['negationInd']
+          is_reason = false
```
okeefm: Don't need this here; `nil` evaluates to `false` so if `is_reason` doesn't get set, it'll just not evaluate to true below in the if statement.

lib/health-data-standards/import/cat1/patient_importer.rb
```diff
@@ -9,6 +9,7 @@ module Cat1
       # This class is a Singleton. It should be accessed by calling PatientImporter.instance
       class PatientImporter
         include Singleton
+        include HealthDataStandards::Util
```
okeefm: What is getting included by this? Can we scope it more narrowly, so we're not including all the classes in HDS::Util?

lib/health-data-standards/export/helper/scooped_view_helper.rb
```diff
@@ -141,12 +146,12 @@ def entries_for_data_criteria(data_criteria, patient)
...
                 if (entry.transferFrom)
                   entry.transferFrom.codes[entry.transferFrom.code_system] = [entry.transferFrom.code]
-                  !entry.transferFrom.codes_in_code_set(codes).empty?
+                  entry.transferFrom.codes_in_code_set(codes).values.first.size > 0
                 end
               elsif data_criteria_oid == '2.16.840.1.113883.3.560.1.72'
                 if (entry.transferTo)
                   entry.transferTo.codes[entry.transferTo.code_system] = [entry.transferTo.code]
-                  !entry.transferTo.codes_in_code_set(codes).empty?
+                  entry.transferTo.codes_in_code_set(codes).values.first.size > 0
```
okeefm: Is there a reason we're not doing something like `!entry.transferTo.codes_in_code_set(codes).values.first.empty?` any more?

dczulada: it was returning a hash {"CODE_SYSTEM" => []}.  this was never registering as empty

okeefm: If `entry.transferTo.codes_in_code_set(codes).values.first` is returning `{"CODE_SYSTEM" => []}`, shouldn't the size of that be 1, though?

dczulada: The size is 0 for an empty array

okeefm:
@dczulada can you fix the merge conflicts on this branch by rebasing? Once you do that I can pull it in.

PR #364, opened 2016-02-24
importer and exporters for device order not done workaround

dczulada:


lib/health-data-standards/import/cat1/patient_importer.rb
```diff
@@ -51,7 +51,8 @@ def initialize
                                             generate_importer(CDA::AllergyImporter, "./cda:entry/cda:observation[cda:templateId/@root = '2.16.840.1.113883.10.20.24.3.43']", '2.16.840.1.113883.3.560.1.7'), #medication adverse effect
                                             generate_importer(CDA::AllergyImporter, "./cda:entry/cda:observation[cda:templateId/@root = '2.16.840.1.113883.10.20.24.3.44']", '2.16.840.1.113883.3.560.1.1')] #medication allergy
 
-          @section_importers[:medical_equipment] = [generate_importer(CDA::MedicalEquipmentImporter, "./cda:entry/cda:procedure[cda:templateId/@root = '2.16.840.1.113883.10.20.24.3.7']", '2.16.840.1.113883.3.560.1.110', 'applied')]
+          @section_importers[:medical_equipment] = [generate_importer(CDA::MedicalEquipmentImporter, "./cda:entry/cda:procedure[cda:templateId/@root = '2.16.840.1.113883.10.20.24.3.7']", '2.16.840.1.113883.3.560.1.110', 'applied'),
+                                                    generate_importer(CDA::MedicalEquipmentImporter, "./cda:entry/cda:act[cda:code/@code = 'SPLY']", '2.16.840.1.113883.3.560.1.137')]
```
rdingwell, line 55: Has it been verified that the only items that meet this xpath "./cda:entry/cda:act[cda:code/@code = 'SPLY']"  will be the workarounds?  Are there any other qrda templates that have an act with a code == sply ?  Also feel a bit uneasy about the fact that there is no template id for the negated template.

dczulada, line 55: As far as I can see in our templates, this would be the only item that would meet the xpath.
I'm uneasy as well...but there isn't an official IG for this template, so i don't believe there would be an associated id

PR #370, opened 2016-03-11
Updates for CMS IG compliance

dczulada:


lib/health-data-standards/export/cms_cat_1.rb
```diff
@@ -0,0 +1,19 @@
+module HealthDataStandards
+  module Export
+    class CMSCat1
```
okeefm: Probably could just add the `cms_compatibility` param to the `Cat1` exporter, rather than making a new one.

PR #374, opened 2016-03-16
defaults to HQR_EHR if no measures are passed into generator

dczulada:


ssayer:
I think we need to be careful about defaulting to EP vs EH. If someone chooses to certify to EH and EP measures we can't just default to HQR. Also, we shouldn't be generating a QRDA file if there are no measures

PR #382, opened 2016-05-04
Added principal diagnosis qrda export for encounters

holmesie:


templates/cat1/_2.16.840.1.113883.10.20.24.3.23.cat1.erb
```diff
@@ -93,5 +93,14 @@
     <% if entry.reason.present? -%>
       <%== render(:partial => 'reason', :locals => {:entry => entry,:reason_oids=>field_oids["REASON"]}) %>
     <% end -%>
+
+    <% if defined?(entry['principalDiagnosis']) -%>
```
dczulada: This should just be

<% if entry.principalDiagnosis -%>

PR #383, opened 2016-05-16
add categories to vs

dczulada:


lib/health-data-standards/models/svs/value_set.rb
```diff
@@ -40,17 +41,34 @@ def self.load_from_xml(doc)
...
+                    display_name: con["displayName"], code_system: con["codeSystem"])
+        end
+      end
+
+      def self.extract_categories(vs_element)
+        category_hash = Hash.new {|h,k| h[k]=[]}
+        groups_with_categories = vs_element.xpath("//vs:Group/@ID[../@displayName='CATEGORY']")
+        groups_with_categories.each do |group_number|
+          measure = vs_element.xpath("//vs:Group[@displayName='CMS eMeasure ID' and @ID='#{group_number}']/vs:Keyword").text
+          categories_for_group = vs_element.xpath("//vs:Group[@displayName='CATEGORY' and @ID='#{group_number}']/vs:Keyword").collect
```
pkmitre: I don't think the "collect" method call at the end of this line is needed, the result of the xpath call should be enumerable without it.

PR #390, opened 2016-06-13
Qrda3 1

dczulada:


mulcahyk:
So far I haven't found anything that differs (significantly) from my review of the updated standard's implementation guide.

I did see that the reason ActRelationshipType was still there for a number of oids.  After talking with David is appears that having these won't be an issue.

PR #392, opened 2016-06-24
Updated default VSAC profile.

hossenlopp:
- VSAC changed the name of the default profile. This causes requests using that profile to return with 404.
- Updated unit test for default VSAC Profile change.


okeefm (approved):
Looks good to me. Just needs to be rebased.

PR #395, opened 2016-07-07
include valuesets with targetSiteCode, dischargeDisposition, facility…

dczulada:
…Location and medicationRoute

### DIFF
--- a/templates/cat1/_2.16.840.1.113883.10.20.24.3.105.cat1.erb
+++ b/templates/cat1/_2.16.840.1.113883.10.20.24.3.105.cat1.erb
@@ -33,7 +33,10 @@
         <% end %>
         <% if entry.route.present? -%>
         <!-- Attribute: route -->
-        <routeCode code="<%= entry.route['code'] %>" codeSystem="<%= HealthDataStandards::Util::CodeSystemHelper.oid_for_code_system(entry.route['code_system']) %>"/>
+          <% vset = oid_for_code(entry.route.code,field_oids["ROUTE"]) %>
+          <% if vset %>
+            <routeCode code="<%= entry.route['code'] %>" codeSystem="<%= HealthDataStandards::Util::CodeSystemHelper.oid_for_code_system(entry.route['code_system']) %>" sdtc:valueSet="<%= vset %>"/>
+          <% end %>
         <% end -%>
         
         <!-- Attribute: dose -->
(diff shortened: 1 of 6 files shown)


dczulada:
https://jira.oncprojectracking.org/browse/CYPRESS-674?

templates/cat1/_2.16.840.1.113883.10.20.24.3.105.cat1.erb
```diff
@@ -33,7 +33,10 @@
         <% end %>
         <% if entry.route.present? -%>
         <!-- Attribute: route -->
-        <routeCode code="<%= entry.route['code'] %>" codeSystem="<%= HealthDataStandards::Util::CodeSystemHelper.oid_for_code_system(entry.route['code_system']) %>"/>
+          <% vset = oid_for_code(entry.route.code,field_oids["ROUTE"]) %>
+          <% if vset %>
+            <routeCode code="<%= entry.route['code'] %>" codeSystem="<%= HealthDataStandards::Util::CodeSystemHelper.oid_for_code_system(entry.route['code_system']) %>" sdtc:valueSet="<%= vset %>"/>
```
edeyoung, line 38: it looks like `sdtc:valueSet` is a "SHOULD" rather than an "SHALL". Should the `<% if vset %>` be moved to be around the attribute rather than around the entire element?

~~Also, how should this work in R2? I don't see a `valueSet` mentioned in R2 so should there also be a `r2_compatibility` check in here somewhere?~~
EDIT: I think I see that it's available in R2 so this shouldn't effect R2 compatibility

templates/cat1/_2.16.840.1.113883.10.20.24.3.23.cat1.erb
```diff
@@ -14,9 +14,12 @@
...
       <high <%= value_or_null_flavor(entry.discharge_time || entry.end_time) %>/>
     </effectiveTime>
-      <% if entry.discharge_disposition.present? -%>
-    <sdtc:dischargeDispositionCode code="<%= entry.discharge_disposition['code'] %>" 
-            codeSystem="<%= HealthDataStandards::Util::CodeSystemHelper.oid_for_code_system(entry.discharge_disposition['codeSystem'] || entry.discharge_disposition['code_system']) %>"/>
+      <% if entry.discharge_disposition.present?
+      vset = oid_for_code(entry.discharge_disposition,field_oids["DISCHARGE_STATUS"])
+      if vset %>  
+        <sdtc:dischargeDispositionCode code="<%= entry.discharge_disposition['code'] %>" 
+            codeSystem="<%= HealthDataStandards::Util::CodeSystemHelper.oid_for_code_system(entry.discharge_disposition['codeSystem'] || entry.discharge_disposition['code_system']) %>" sdtc:valueSet="<%= vset %>"/>
```
edeyoung, line 21: I can't find a reference to `sdtc:valueSet` in the spec for `dischargeDispositionCode`. Is this referenced somewhere else in the documentation?

Also, if it's a "SHOULD" rather than a "SHALL" (like above), consider changing the position of the `if vset` check here as well.

edeyoung, line 21: (similar comments for the rest of the changes in this pull request)

PR #396, opened 2016-07-07
Types of PQ need units.  non-units in UCUM can be 1 or {word}

dczulada:


ssayer:
Is there a reason you moved this logic into the ERB file and didn't just update the helper

dczulada:
I didn't move any logic here, just updated in two places

ssayer:
Oh my bad, I read that first diff backwards for some reason

lib/health-data-standards/export/view_helper.rb
```diff
@@ -53,7 +53,7 @@ def status_code_for(entry)
       def fulfillment_quantity(codes, fulfillmentHistory, dose)
         if (codes["RxNorm"].present?)
           doses = (fulfillmentHistory.quantity_dispensed['value'].to_f / dose['value'].to_f ).to_i
-          return "value='#{doses}'"
+          return "value='#{doses}' unit='1'"
```
edeyoung, line 56: Based on the commit message, it sounds like something that doesn't have units is valid if you include `unit='1'` (maybe?). If that's the case, a comment here would be helpful because this return string is unintuitive.

templates/cat1/_result_value.cat1.erb
```diff
@@ -14,7 +14,7 @@
 	<%== code_display(ev, 'preferred_code_sets' => ['LOINC', 'SNOMED-CT', 'ICD-9-CM', 'ICD-10-CM'], 'tag_name' => 'value', 'extra_content' => extra_content) %>
 	<%   elsif ev.respond_to?(:scalar) -%>
 	<%     if is_num?(ev.scalar) -%>
-	<value xsi:type="PQ" value="<%= ev.scalar %>" <% if ev.units && (ev.units != "") -%>unit="<%= ev.units %>"<% end -%>/>
+	<value xsi:type="PQ" value="<%= ev.scalar %>" <% if ev.units && (ev.units != "") -%>unit="<%= ev.units %>"<% else -%>unit="1"<% end -%>/>
```
edeyoung, line 17: A comment here would also be helpful to explain the role of `unit='1'`

dczulada:
agree with having comment to make it more clear.

PR #398, opened 2016-07-07
Fix CMS 2016 IG incompatibility in record target

tstrass:
Was getting this warning in each qrda file:

screen shot 2016-07-06 at 9 25 48 am


ssayer:
If we have a MRN/MRA hardcoded even though it's present, it's going to cause problems because patients will look like duplicates

okeefm:
The CMS IG expects there to be two `id` elements in that section. However, one should have the root of `2.16.840.1.113883.4.572`, and represent the Medicare HIC number, and the other should be the MRN/MRA. See page 13 of the 2016 CMS IG for more details.

ssayer:
Yeah, I just want to make sure it's not going to bork any code dealing with duplicates by having two ID elements in there.

PR #400, opened 2016-07-11
added HCPCS and CDT to preferred encounter code sets

dczulada:


templates/cat1/_2.16.840.1.113883.10.20.24.3.22.cat1.erb
```diff
@@ -5,7 +5,7 @@
     <!-- Encounter order template -->
     <templateId root="2.16.840.1.113883.10.20.24.3.22" <% if !r2_compatibility %>extension="2014-12-01"<%end%>/>
     <id root="1.3.6.1.4.1.115" extension="<%= entry.id %>"/>
-    <%== code_display(entry, 'preferred_code_sets' => ['SNOMED-CT', 'ICD-9-CM', 'ICD-10-CM', 'CPT'], 'value_set_map' => filtered_vs_map, 'extra_content' => "sdtc:valueSet=\"#{value_set_oid}\"") %>
+    <%== code_display(entry, 'preferred_code_sets' => ['SNOMED-CT', 'ICD-9-CM', 'ICD-10-CM', 'CPT', 'HCPCS', 'CDT'], 'value_set_map' => filtered_vs_map, 'extra_content' => "sdtc:valueSet=\"#{value_set_oid}\"") %>
```
edeyoung, line 8: Are these preferred codes listed in the docs? If not, this should probably be changed to `'preferred_code_sets' => ['*']` so we don't accidentally kick out some code sets.

Same comment for the rest of the file.

edeyoung, line 8: Also, `preferred_code_sets` are found in a bunch of templates and seem to refer to restrictions in R2 rather than R3 or R3.1. On Bonnie, we recently had an issue with a user and his code set getting kicked out due to an outdated restriction. I think it would be worth going through the templates and updating these restrictions to match the specs. (Bonnie team is happy to do this (along with several of the other suggestions - let's coordinate)).

dczulada:
I’m ok with the approach of moving to [‘*']

PR #401, opened 2016-07-15
Old tobacco use template was removed from QRDA r3

dczulada:


lib/health-data-standards/export/qrda/entry_template_resolver.rb
```diff
@@ -53,11 +53,18 @@ def qrda_oid_for_hqmf_patient_characteristic(vs_oid, qrda_version = nil)
             '2.16.840.1.113883.10.20.24.3.101'
           when '2.16.840.1.113883.3.526.3.1189', '2.16.840.1.113883.3.526.3.1170', '2.16.840.1.113883.3.600.2390'
             # Patient Characteristic Tobacco User/Non-User
-            '2.16.840.1.113883.10.20.22.4.85'
+            case qrda_version
```
edeyoung, line 56: should these `case` statements be more like:

```
when 'r3_1'
  ...
when 'r3'
  ...
when 'r2'
  ...
else
  < some sort of exception for an unsupported version >
```

As written, we default to `r2`. If we were to update to `r4` in the future but forget to update this `case` statement, we would fall through and get the `r2` template. This also fails silently if someone tries to pass in an unsupported version.

templates/cat1/show.cat1.erb
```diff
@@ -171,7 +171,7 @@
     <structuredBody>
       <%== render :partial => 'measures', :locals => {:measures => measures, :patient => patient} %>
       <%== render :partial => 'reporting_parameters', :locals => {:start_date => start_date, :end_date => end_date, :cms_compatibility => cms_compatibility} %>
-      <%== render :partial => 'patient_data', :locals => {:measures => measures, :patient => patient, :r2_compatibility => r2_compatibility, :cms_compatibility => cms_compatibility} %>
+      <%== render :partial => 'patient_data', :locals => {:measures => measures, :patient => patient, :r2_compatibility => r2_compatibility, :cms_compatibility => cms_compatibility, :qrda_version => qrda_version} %>
```
edeyoung, line 174: It seems like `r2_compatibility` and `qrda_version` are trying to (somewhat) represent the same things with the limitation that `r2_compatibility` can only represent "is R2" and "is not R2". But they're both giving information about the version.

What about changing the code (this would be a pain, but it would reduce confusion, increase readability, and probably limit future errors) so all that's used is `qrda_version`?

lib/health-data-standards/export/qrda/entry_template_resolver.rb
dczulada, line 56: case statement looks good

templates/cat1/show.cat1.erb
dczulada, line 174: Yes, r2_compatibility is similar to qrda_version.  This was left in place to make as few changes as possible to the legacy code.  Moving forward, if we want to still support QRDA Cat I R2 we should use the same construct as the other versions.

PR #402, opened 2016-07-18
HQMF parser now preserves the full HQMF eMeasure version number

holmesie:


mulcahyk:
Needs to have the version number for the measure in Bonnie (actually bonnie_bundler) updated from int to string.

PR #407, opened 2016-07-25
add code systems used in measures into the preferred_code_sets

dczulada:


templates/cat1/_2.16.840.1.113883.10.20.24.3.3.cat1.erb
```diff
@@ -2,7 +2,7 @@
   <act classCode="ACT" moodCode="EVN" <%== negation_indicator(entry) %>>
     <templateId root="2.16.840.1.113883.10.20.24.3.3" <% if !r2_compatibility %>extension="2014-12-01"<%end%>/>
     <id root="1.3.6.1.4.1.115" extension="<%= entry.id %>"/>  
-    <%== code_display(entry, 'value_set_map' => filtered_vs_map, 'preferred_code_sets' => ['SNOMED-CT'], 'extra_content' => "sdtc:valueSet=\"#{value_set_oid}\"") %>
+    <%== code_display(entry, 'value_set_map' => filtered_vs_map, 'preferred_code_sets' => ['SNOMED-CT', 'LOINC'], 'extra_content' => "sdtc:valueSet=\"#{value_set_oid}\"") %>
```
edeyoung, line 5: I mentioned this in a previous pull request, but I think it's worth going through the docs and double checking if `preferred_code_sets` are required for each template and replacing the array with `['*']` if they aren't.

dczulada, line 5: I’m ok with the approach of moving to [‘*']

PR #416, opened 2016-07-27
add support for QRDA Cat III R1.1

dczulada:


templates/cat3/_measure_data.cat3.erb
```diff
@@ -2,7 +2,7 @@
 <!--   MEASURE DATA REPORTING FOR    <%= population.type %>  <%= population.id %>  -->
 <observation classCode="OBS" moodCode="EVN">
   <!-- Measure Data template -->
-  <templateId root="2.16.840.1.113883.10.20.27.3.5"/>
+  <templateId root="2.16.840.1.113883.10.20.27.3.5" <% if qrda3_version == 'r1_1' %>extension="2016-02-01"<%end%>/>
```
edeyoung, line 5: this seems to follow the same method as all the `r2_compatibility` flags in the QRDA Cat 1 templates, which was unscalable as new versions were released. Consider either doing the file name extensions with the date like what was done for r3.1 cat 1 QRDA templates or place all templates related to the new version in a folder for that version. (e.g. you would have `templates/cat3/r3_1/_measure_data.cat3.erb`).
(I know this is all switching to Go, so maybe making this work with future versions is not a big concern here.)

lib/health-data-standards/export/cat_3.rb
```diff
@@ -9,15 +9,15 @@ def initialize
         @cat1_renderer.template_helper = HealthDataStandards::Export::TemplateHelper.new('cat1', 'cat1')
       end
 
-      def export(measures, header, effective_date, start_date, end_date, filter=nil,test_id=nil)
+      def export(measures, header, effective_date, start_date, end_date, qrda3_version=nil, filter=nil,test_id=nil)
```
edeyoung, line 12: the `qrda3_version` flag appears to be named specifically for minor revisions of QRDA R3. Consider renaming to just `qrda_version` so that it will easily fit for future major versions of QRDA.

dczulada, line 12: Should probably be “qrda_cat_3_version”. (existing qrda_version should probably be “qrda_cat_1_version”).  Would probably make sense to break out versions in the same was as QRDA 1 (thou this would be the longer term).

PR #420, opened 2016-08-10
add missing parameter to build_error call

dehall:
The build_error function takes 3 parameters: the message, the location (xpath), and options. In this case the file_errors do not provide an xpath so our choices are to either use a simple default (the root `/`) or put the line number in the location field. Putting the line number in the location field would make everyone have to deal with this new format so choosing the default for now

### DIFF
--- a/lib/health-data-standards/validate/schematron_validator.rb
+++ b/lib/health-data-standards/validate/schematron_validator.rb
@@ -25,7 +25,7 @@ module HealthDataStandards
           file_errors = document.errors.select { |e| e.fatal? || e.error? }
           if file_errors
             file_errors.each do |error|
-              build_error error, :file_name => data[:file_name] ,:validator => :xml_validation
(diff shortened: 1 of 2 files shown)


lib/health-data-standards/validate/schematron_validator.rb
```diff
@@ -25,7 +25,7 @@ def validate(document,data = {})
           file_errors = document.errors.select { |e| e.fatal? || e.error? }
           if file_errors
             file_errors.each do |error|
-              build_error error, :file_name => data[:file_name] ,:validator => :xml_validation
+              build_error error, '/', :file_name => data[:file_name] ,:validator => :xml_validation
```
okeefm: Is there a reason we use named parameters for `:file_name` and `:validator`? I'm not even sure `:validator` does anything in this context. The `build_error` method looks like this: 

``````
def build_error(msg, loc, file_name)
    ValidationError.new(message: msg, location: loc, file_name: file_name, validator: @name)
end```
``````

PR #421, opened 2016-08-18
Made "Medication, Order" cumulative medication duration calculate correctly. 

edeyoung:
This addresses https://oncprojectracking.healthit.gov/support/browse/BONNIE-172.

Usually, cumulative medication duration is calculated based off of the fulfillment history, dose, and regimen. However, for "Medication, Order", the medication has not yet been filled, so fulfillment history and dose don't make sense. Maybe more importantly, these fields are not available in the QRDA for the "Medication, Order" template.

Instead, cumulative medication duration should be calculated based on the maximum allowed administrations (`repeatNumber` in the QRDA) and the regimen (`effectiveTime` with `period` subelement in the QRDA).

Additionally, fulfillment information should not be allowed to be entered for "Medication, Order".

Specific changes for this repository include adding `allowedAdministrations` to the medication model and including `repeateNumber` in the "Medication, Ordered" QRDA templates.

This pull request depends on the `medication_order_cmd` branches in **bonnie, hqmf2js, and patient-api.**

### DIFF
--- a/templates/cat1/_2.16.840.1.113883.10.20.24.3.47.cat1.erb
+++ b/templates/cat1/_2.16.840.1.113883.10.20.24.3.47.cat1.erb
@@ -21,6 +21,9 @@
         <period value="<%= period['value']%>" unit="<%= period['unit']%>"/>
       </effectiveTime>
     <% end %>
+    <% if entry.allowedAdministrations.present? %>
+      <repeatNumber value="<%== entry.allowedAdministrations %>" />
+    <% end %>
      <%== render(:partial => 'medication_details', :locals => {:entry => entry, :route_oids=>field_oids["ROUTE"]}) %>
     <consumable>
       <manufacturedProduct classCode="MANU">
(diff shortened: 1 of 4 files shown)


templates/cat1/_2.16.840.1.113883.10.20.24.3.47.cat1.erb
```diff
@@ -21,6 +21,9 @@
         <period value="<%= period['value']%>" unit="<%= period['unit']%>"/>
       </effectiveTime>
     <% end %>
+    <% if entry.allowedAdministrations.present? %>
+      <repeatNumber value="<%== entry.allowedAdministrations %>" />
```
okeefm, line 25: Do we want to be able to import this field too? If so, we should add something to the `CDA::MedicationImporter` (which imports this template currently), or subclass it and create a `Cat1::MedicationImporter`

okeefm (changes requested):

thecristen (changes requested):
I think mokeefe's comment should be addressed, other than that I'd say we're good to go.

thecristen (approved):

okeefm (approved):
Changes look good to me.

PR #423, opened 2016-08-23
Display name was incorrect for 2.16.840.1.113883.10.20.24.3.82

dczulada:
TransferFrom and TransferTo ids are not randomized when patients are cloned, when the id is exported, use the hash of the transfer and encounter id

### DIFF
--- a/templates/cat1/_2.16.840.1.113883.10.20.24.3.81.cat1.erb
+++ b/templates/cat1/_2.16.840.1.113883.10.20.24.3.81.cat1.erb
@@ -1,7 +1,7 @@
 <entry>
 	<encounter classCode="ENC" moodCode="EVN">
 		<templateId root="2.16.840.1.113883.10.20.24.3.81" <% if !r2_compatibility %>extension="2014-12-01"<%end%>/>
-		<id root="1.3.6.1.4.1.115" extension="<%= entry.transferFrom.id %>"/>
+    	<id root="1.3.6.1.4.1.115" extension="<%= "#{entry.transferFrom.id}#{entry.id}".hash.abs.to_s(16) %>"/>
 		<code code="77305-1" codeSystem="2.16.840.1.113883.6.1" codeSystemName="LOINC" displayName="transferred from"/>
 		<statusCode code="completed"/>
 		<participant typeCode="ORG">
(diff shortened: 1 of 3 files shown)


templates/cat1/_2.16.840.1.113883.10.20.24.3.81.cat1.erb
```diff
@@ -1,7 +1,7 @@
 <entry>
 	<encounter classCode="ENC" moodCode="EVN">
 		<templateId root="2.16.840.1.113883.10.20.24.3.81" <% if !r2_compatibility %>extension="2014-12-01"<%end%>/>
-		<id root="1.3.6.1.4.1.115" extension="<%= entry.transferFrom.id %>"/>
+    	<id root="1.3.6.1.4.1.115" extension="<%= "#{entry.transferFrom.id}#{entry.id}".hash.abs.to_s(16) %>"/>
```
holmesie: It looks like your editor might be replacing tabs with spaces here, might be worth keeping these as tabs for consistency.

templates/cat1/_2.16.840.1.113883.10.20.24.3.81_2016_02_01.cat1.erb
```diff
@@ -1,7 +1,7 @@
 <entry>
 	<encounter classCode="ENC" moodCode="EVN">
 		<templateId root="2.16.840.1.113883.10.20.24.3.81" extension="2016-02-01"/>
-		<id root="1.3.6.1.4.1.115" extension="<%= entry.transferFrom.id %>"/>
+    	<id root="1.3.6.1.4.1.115" extension="<%= "#{entry.transferFrom.id}#{entry.id}".hash.abs.to_s(16) %>"/>
```
holmesie: See previous comment about tabs => spaces.

holmesie:
It looks like **_2.16.840.1.113883.10.20.24.3.82_2016_02_01.cat1.erb** is essentially the same thing as **_2.16.840.1.113883.10.20.24.3.82.cat1.erb**, minus the templateId extension. It looks like that might be the same case for **_2.16.840.1.113883.10.20.24.3.81.cat1.erb** and **_2.16.840.1.113883.10.20.24.3.81_2016_02_01.cat1.erb** - do you think it would be possible to combine these somehow?

dczulada:
Ordinarily i would say yes to consolidation.  but moving forward, we will likely have one erb template for every version of every qrda template

PR #425, opened 2016-08-24
QRDA R3.1 allows ordinality in priorityCode or a nested observation.

dczulada:


lib/health-data-standards/import/cda/procedure_importer.rb
```diff
@@ -9,7 +9,7 @@ def initialize(entry_finder=EntryFinder.new("//cda:section[cda:templateId/@root!
           super(entry_finder)
           @entry_class = Procedure
           @value_xpath = "./cda:value | ./cda:entryRelationship[@typeCode='REFR']/cda:observation/cda:value"
-          @ordinality_xpath = "./cda:priorityCode"
+          @ordinality_xpath = "./cda:priorityCode | ./cda:entryRelationship/cda:observation[./cda:code[@code='260870009']]/cda:value"
```
okeefm: Should this code be implemented in the `cat1/procedure_performed_importer.rb`, rather than the `cda/procedure_importer.rb`, since it's specific to QRDA Cat I's procedure performed?

dczulada: yes

PR #431, opened 2016-09-23
added the 'method' attribute onto assessments and added a result path…

edeyoung:
… to the assessment performed oid.

See pull requests in bonnie and patientapi as well.

See attached measures to test the added attributes.

*\* assessment performed **

method
result
negatable

*\* assessment recommended **

method
negatable

### DIFF
--- a/lib/health-data-standards/models/assessment.rb
+++ b/lib/health-data-standards/models/assessment.rb
@@ -1,2 +1,3 @@
 class Assessment < Entry
+  field :method,              type: Hash
 end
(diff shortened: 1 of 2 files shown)


lib/health-data-standards/models/assessment.rb
```diff
@@ -1,2 +1,3 @@
 class Assessment < Entry
+  field :method,              type: Hash
```
holmesie, line 2: Were all these extra spaces intentional?

edeyoung: a lot of the other files have multiple of these lists and the spaces create nice columns. Here there's only one so it's a bit weird looking. I can see an argument for keeping or getting rid of them. Let me know what you prefer.

holmesie: Ah that makes sense - I think I prefer keeping things consistent, so it's good as-is (in my opinion).

PR #433, opened 2016-10-04
QRDA Cat III v1.1 uses IPOP instead of IPP all other population codes…

dczulada:
… are the same

### DIFF
--- a/templates/cat3/_measure_data.cat3.erb
+++ b/templates/cat3/_measure_data.cat3.erb
@@ -121,9 +121,10 @@
        if payer_supplimental_data.present?
          payer_supplimental_data.each do |payer, count| -%>
 
- <!--         PAYER Supplemental Data Reporting   for<%= population.type %>  <%= population.id %>   -->        
+ <!--         PAYER Supplemental Data Reporting   for<%= population.type %>  <%= population.id %>   -->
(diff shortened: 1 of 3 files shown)


templates/cat3/_measure_data.cat3.erb
```diff
@@ -8,9 +8,15 @@
         displayName="Assertion" 
         codeSystemName="ActCode"/>
   <statusCode code="completed"/>
-  <value xsi:type="CD" code="<%= population.type %>" 
-         codeSystem="2.16.840.1.113883.5.1063"  
-         codeSystemName="ObservationValue"/>
+  <% if qrda3_version == 'r1_1' && population.type == 'IPP' %>
+    <value xsi:type="CD" code="IPOP"
```
okeefm: Is this code actually in the `ObservationValue` Code system? I can't find it on the [HL7 page for this Code System](http://www.hl7.org/documentcenter/public_temp_A99763EF-1C23-BA17-0CB4A1C906C89669/standards/vocabulary/vocabulary_tables/infrastructure/vocabulary/ObservationValue.html)

okeefm: Looks like this may be the new IPOP code:

```
<value xsi:type="CD" code="IPOP"
         codeSystem="2.16.840.1.113883.5.4"
         displayName="initial population"
         codeSystemName="ActCode"/>
```

Based on the QRDA 3 IG. I think this should be the code system for all the populations in QRDA 3 (see pages 71 and 72 of the QRDA 3 IG).

okeefm (changes requested):

okeefm (approved):
Only comment: do we want to include a display name? If not, that's fine too. Changes to population values/codesystem look good to me

PR #435, opened 2016-10-14
Characteristic not Charactersitic

dczulada:


okeefm (changes requested):
There are 3 other places `Charactersitic` needs to be replaced with `Characteristic` in this file

okeefm (approved):
Looks good to me.

PR #438, opened 2016-10-25
Qrda 3.1 validation

alexanderelliott121:
Change the way the qrda version is passed through, how templates are stored and template logic surrounding qrda version.


lib/health-data-standards/export/qrda/entry_template_resolver.rb
```diff
@@ -46,7 +46,7 @@ def qrda_oid_for_hqmf_patient_characteristic(vs_oid, qrda_version = nil)
           when '2.16.840.1.113883.3.526.3.1279'
             # Patient Characteristic Observation Assertion template for
             # Patient Characteristic: ECOG Performance Status-Poor
-            qrda_version == 'r3_1' ? '2.16.840.1.113883.10.20.24.3.103_2016_02_01' : '2.16.840.1.113883.10.20.24.3.103'
+            qrda_version == 'r3_1' ? '2.16.840.1.113883.10.20.24.3.103' : '2.16.840.1.113883.10.20.24.3.103'
```
edeyoung: are these the same ids in either case now? if so, can we remove the `...?...:...` stuff and just return the one oid?

lib/health-data-standards/export/qrda/entry_template_resolver.rb
```diff
@@ -55,15 +55,15 @@ def qrda_oid_for_hqmf_patient_characteristic(vs_oid, qrda_version = nil)
             # Patient Characteristic Tobacco User/Non-User
             case qrda_version
             when 'r3_1'
-              '2.16.840.1.113883.10.20.24.3.103_2016_02_01'
+              '2.16.840.1.113883.10.20.24.3.103'
```
edeyoung: can you have `or`s in `when` statements? If so, combine `when 'r3_1'` and `when 'r3'`.

lib/health-data-standards/export/qrda/entry_template_resolver.rb
```diff
@@ -55,15 +55,15 @@ def qrda_oid_for_hqmf_patient_characteristic(vs_oid, qrda_version = nil)
...
             when 'r3'
               '2.16.840.1.113883.10.20.24.3.103'
             else
               '2.16.840.1.113883.10.20.22.4.85'
             end
           else
             # return generic pc observation template for anything not specificly mapped to its own template
-            qrda_version == 'r3_1' ? '2.16.840.1.113883.10.20.24.3.103_2016_02_01' : '2.16.840.1.113883.10.20.24.3.103'
+            qrda_version == 'r3_1' ? '2.16.840.1.113883.10.20.24.3.103' : '2.16.840.1.113883.10.20.24.3.103'
```
edeyoung: See above. you can get rid of the `...?...:...` stuff here.

lib/health-data-standards/export/cat_1.rb
```diff
@@ -1,30 +1,23 @@
...
-          qrda_template = 'show'
-          r2_compatibility = false
-        elsif qrda_version == 'r2'
-          qrda_template = 'show'
+        qrda_template = 'show'
+        r2_compatibility = false
+
+        if qrda_version == 'r2'
           r2_compatibility = true
-        # defaults to QRDA R3 'show' templates (prior to version specific templates)
```
edeyoung, line 24: what happens if the `qrda_version` isn't specified? Does this fail gracefully? There are two situations we should consider:
- `nil` is passed in. In this case, we should consider defaulting to some version (the latest?)
- an unexpected value is passed in. In this case, we should error gracefully with an explanatory error message.

lib/health-data-standards/export/template_helper.rb
```diff
@@ -29,21 +30,26 @@ def template_root
...
+      def partial(partial_name, provided_qrda_version=nil)
+        cache_template("_#{partial_name}", provided_qrda_version)
       end
 
       protected 
 
-      def cache_template(template_name)
+      def cache_template(template_name, provided_qrda_version=nil)
+        subdir = @qrda_version
+        subdir = provided_qrda_version if provided_qrda_version
```
edeyoung: should there be some sort of error if the qrda versions don't align?

does it make sense to only specify the qrda version in one place - either in the template helper or the renderer (currently it can be passed in in either) so this kind of conflict can't occur?

templates/cat1/r2/_2.16.840.1.113883.10.20.24.3.9.cat1.erb
```diff
@@ -30,7 +30,11 @@
     <%== render(:partial => 'reason', :locals => {:entry => entry, :reason_oids=>field_oids["REASON"], :r2_compatibility => r2_compatibility}) %>
   </supply>
   <% else %>
-  <act classCode="ACT" moodCode="RQO" <%== negation_indicator(entry) %>>
+  <act classCode="SPLY" moodCode="RQO" <%== negation_indicator(entry) %>>
+    <!-- Plan of Care Activity Supply -->
+    <templateId root="2.16.840.1.113883.10.20.22.4.43" <% if !r2_compatibility %>extension="2014-06-09"<%end%>/>
```
edeyoung, line 35: is the plan still to go back through the `r2` templates and remove the `r2_compatibility` flags? These aren't needed anymore. I remember we discussed removing them and decided that in the short term we would just do `r3` and `r3_1`. Am I remembering correctly?

templates/cat1/r3/_2.16.840.1.113883.10.20.24.3.9.cat1.erb
```diff
@@ -0,0 +1,64 @@
...
+          <%== code_display(entry, 'value_set_map' => filtered_vs_map, 'preferred_code_sets' => ['LOINC', 'SNOMED-CT', 'ICD-9-PCS', 'ICD-10-PCS', 'CPT'], 'extra_content' => "sdtc:valueSet=\"#{value_set_oid}\"") %>
+        </playingDevice>
+      </participantRole>
+    </participant>
+    <%== render(:partial => 'reason', :locals => {:entry => entry, :reason_oids=>field_oids["REASON"]}) %>
+  </supply>
+  <% else %>
+  <act classCode="SPLY" moodCode="RQO" <%== negation_indicator(entry) %>>
+    <!-- Plan of Care Activity Supply -->
+    <templateId root="2.16.840.1.113883.10.20.22.4.43" <% if !r2_compatibility %>extension="2014-06-09"<%end%>/>
```
edeyoung: are these `r2_compatibility` flags generally still present in the `r3` templates? is this one of the "future change" things? I had it in my head it was only the `r2` files that weren't going to be changed, but if it's these too, that's fine and good to know.

edeyoung (changes requested):
a key thing to figure out is how this will handle an unspecified or bad `qrda_version`.

edeyoung:
`r3` template `2.16.840.1.113883.10.20.24.3.103.cat1.erb` has the incorrect date associated with it. It has "2016-02-01". It should have "2014-12-01".

lib/health-data-standards/export/qrda/entry_template_resolver.rb
okeefm: You can't have `or`s, but you can basically do the same thing with `when 'r3_1','r3'`

lib/health-data-standards/export/cat_1.rb
okeefm: Could also rewrite this to be a `case` statement similar to other locations in the code, to make it cleaner.

templates/cat1/r3/_2.16.840.1.113883.10.20.24.3.9.cat1.erb
mayerm94: r2_compatibility flags have been removed entirely from the r3 templates

templates/cat1/r2/_2.16.840.1.113883.10.20.24.3.9.cat1.erb
mayerm94: Correct, the r2_compatibility flags will be left in r2 for the time being. However, the r2_compatibility flags have been stripped out of r3 and r3_1

lib/health-data-standards/export/template_helper.rb
alexanderelliott121: The implementation was set up to accommodate preexisting architecture.
I've added a todo to rectify it.

alexanderelliott121:
I've added a check to cat1 that sets a nil qrda_version to the latest.

templates/cat1/r2/show.cat1.erb
```diff
@@ -13,11 +13,11 @@
   <realmCode code="US"/>
   <typeId root="2.16.840.1.113883.1.3" extension="POCD_HD000040"/>
   <!-- US Realm Header Template Id -->
-  <templateId root="2.16.840.1.113883.10.20.22.1.1" <% if !r2_compatibility %>extension="2014-06-09"<%end%> />
+  <templateId root="2.16.840.1.113883.10.20.22.1.1" extension="2014-06-09"/>
   <!-- QRDA templateId -->
-  <templateId root="2.16.840.1.113883.10.20.24.1.1" <% if !r2_compatibility %>extension="2014-12-01"<%end%> />
+  <templateId root="2.16.840.1.113883.10.20.24.1.1" extension="2014-12-01" />
```
okeefm, line 18: I'm not sure I understand why the `if` statement was removed here, but the same `if` statement is present in other cat1 r2 files.

alexanderelliott121: We decided that we would focus on the r3 and r3_1 templates.  That r2 was changed at some point, the rest will be addressed in future work.

edeyoung: i think the problem @okeefm is mentioning is that this template _was_ changed in r2. It looks like this change was in error - it makes the template always include date flags that should not be in an r2 template.

templates/cat1/r3_1/_2.16.840.1.113883.10.20.24.3.23.cat1.erb
```diff
@@ -5,7 +5,7 @@
     <!-- Encounter performed template -->
     <templateId root="2.16.840.1.113883.10.20.24.3.23" extension="2016-02-01"/>
```
dczulada, line 6: In QRDA 3.1 the encounter performed template must be contained in the Encounter Performed Act template to represent the Encounter, Performed QDM data type.

screen shot 2016-11-04 at 12 10 25 pm

lib/health-data-standards/export/qrda/hqmf-qrda-oids.json
```diff
@@ -42,8 +39,7 @@
     "hqmf_name": "Device, Applied",
     "hqmf_oid": "2.16.840.1.113883.3.560.1.10",
     "qrda_name": "Device Applied",
-    "qrda_oid": "2.16.840.1.113883.10.20.24.3.7",
-    "r3_1": "2016_02_01"
+    "qrda_oid": "2.16.840.1.113883.10.20.24.3.7"
   },
   {
     "hqmf_name": "Device, Intolerance",
```
dczulada, line 45: Device, Order and Device, Order not done still have "r3_1": "2016_02_01"

lib/health-data-standards/export/qrda/entry_template_resolver.rb
```diff
@@ -46,24 +46,22 @@ def qrda_oid_for_hqmf_patient_characteristic(vs_oid, qrda_version = nil)
           when '2.16.840.1.113883.3.526.3.1279'
```
dczulada, line 46: "#{'_'+oid_tuple[qrda_version] if oid_tuple[qrda_version]}"  is no longer needed

should only be 

"#{oid_tuple['qrda_oid']}"

edeyoung:
This work has been incorporated into bonnie_master_2016, which has a pull request into HDS master.

PR #441, opened 2016-10-29
changed the clean_hash_id function to take a hashkey argument, which …

edeyoung:
…it assumed it was taking. this made the generated html generate properly.

### DIFF
--- a/lib/health-data-standards/export/view_helper.rb
+++ b/lib/health-data-standards/export/view_helper.rb
@@ -152,7 +152,7 @@ module HealthDataStandards
             clean_hash[codes['codeSystem']] = clean_hash_code_system(codes)
           elsif codes['_id']
             codes.keys.reject {|key| ['_id'].include? key}.each do |hashkey|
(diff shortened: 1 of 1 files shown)


lib/health-data-standards/export/view_helper.rb
```diff
@@ -184,7 +184,7 @@ def clean_hash_code_system(codes)
         end
       end
       
-      def clean_hash_id(codes)
+      def clean_hash_id(codes, hashkey)
```
okeefm, line 187: Is there a technical reason why we're not passing the `hashkey` and the `value` from codes across, rather than passing `codes` in and immediately grabbing the value out of it?

edeyoung: not that i know of. @dczulada, do you know of a reason?

my work was just a bug fix to add a missing argument.

I'm closing this because it's in bonnie_master_2016. should this comment be moved over to that branch?

edeyoung:
this work has been incorporated into bonnie_master_2016, which is a pull request into HDS master.

PR #444, opened 2016-11-10
Do not save the bundle if it did not successfully import

rbclark:
Currently if you import a bundle and it fails because the same version already exists in the database, then the ensure block is still run and the bundle is still saved, leaving 2 records for the same bundle in the database. This adds a check to see if the bundle was ever saved in the first place before running bundle.save again. This issue can be seen now by accessing the cypress bundle import admin page and uploading the same bundle twice.

### DIFF
--- a/lib/health-data-standards/import/bundle/importer.rb
+++ b/lib/health-data-standards/import/bundle/importer.rb
@@ -64,8 +64,11 @@ module HealthDataStandards
 
           return bundle
         ensure
-          bundle.done_importing = true unless bundle.nil?
-          bundle.save
+          # If the bundle is nil or the bundle has never been saved then do not set done_importing or run save.
+          if bundle && bundle.created_at
+            bundle.done_importing = true
+            bundle.save
+          end
         end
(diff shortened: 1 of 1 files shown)


lib/health-data-standards/import/bundle/importer.rb
```diff
@@ -64,8 +64,11 @@ def self.import(zip,  options={})
 
           return bundle
         ensure
-          bundle.done_importing = true unless bundle.nil?
-          bundle.save
+          # If the bundle is nil or the bundle has never been saved then do not set done_importing or run save.
+          unless bundle.nil? || bundle.created_at.nil?
```
okeefm: Could replace this with `if bundle && bundle.created_at`, for those of us who are really bad at reading `unless` clauses. Also gets rid of the use of `.nil?`.

okeefm (changes requested):

okeefm (approved):

edeyoung (approved):

PR #446, opened 2016-11-15
added act wrapper templates to encounter, transfers and medication di…

dczulada:
…spense


dczulada:
These changes would still need to be incorporated into the QRDA 3.1 Branch

okeefm (approved):
Looks good to me.

okeefm:
Question: should we worry about the CodeClimate "similar code" issues? Or the complex method issue?

ssayer:
There might be a better meta-programming way to get rid of the duplicate code, since the only real difference between the new importers is the parent class and the XPath. Whether it's worth the effort to do that, I'm not sure. The initialize method is already a tire fire, you're just adding more tires :)

I'm not sure if the effort to refactor here is worth the time it will take, but these are things to think about/fix in the Go implementation.

PR #448, opened 2016-11-23
validator to check that correct QDM Patient templates are being used

dczulada:


lib/health-data-standards/validate/qrda_qdm_template_validator.rb
```diff
@@ -0,0 +1,188 @@
+module HealthDataStandards
+  module Validate
+    class QrdaQdmTemplateValidator
+      include BaseValidator
+
+      # Hash of templateIds/extensions specified in the Patient Data Section QDM for QRDA R3
```
okeefm, line 6: Is there a better way to do this than a giant hash where 90% of the options are the same string? Would it work to only declare the ones that _aren't_ the same as that string?

dczulada: However, we would still need to have an array of templates that were that string.  All of the templates that have an extension do have the same one.  I don't believe that has to be the case.  There could be a version of QRDA where some templates don't have updates (possibly)

okeefm (approved):
Looks okay to me. I'm not thrilled about the massive hash, but I think it's probably the clearest way to do it.

edeyoung (approved):

PR #449, opened 2016-11-28
2.16.840.1.113883.10.20.22.4.85 is no longer a QRDA template

dczulada:
2.16.840.1.113883.10.20.24.3.105 (in r3) was missing the children templateId extension
2.16.840.1.113883.10.20.24.3.23 encounters no longer include the transfer element, they are in a different entry
2.16.840.1.113883.10.20.24.3.104 typo in the templateId
2.16.840.1.113883.10.20.24.3.9 device order, not done workaround does not have atemplateid


templates/cat1/r3/_2.16.840.1.113883.10.20.24.3.23.cat1.erb
```diff
@@ -61,24 +61,6 @@
     </participant>
     <% end -%>
 
-    <% if entry.respond_to?(:transferTo) && entry.transferTo.present? -%>
```
okeefm, line 64: What happened to this information in R3? Is it just gone, or is it included in another entry now?

dczulada: It is now an entry/template of its own TransferTo = 2.16.840.1.113883.10.20.24.3.82:2014-12-01 and TransferFrom = 2.16.840.1.113883.10.20.24.3.81:2014-12-01

edeyoung (approved):

PR #450, opened 2016-11-29
Bonnie master 2016

edeyoung:
Primarily includes QRDA 3.1 updates where different QRDA versions are split into separate folders.
Additionally includes some added attributes/data types to address user issues and QDM 4.3 updates.


dczulada:
Code climate appears to be angry about 'entry'.  The only change to entry was the addition of the "health_record_field."  In the grand scheme of things this doesn't really seem like it has made entry appreciably worse.  Thou, what is a 'health record field'?

okeefm:
It looks like adding `health_record_field` was just the straw that broke the camel's back in `Entry`. It had an "A" rating before, and has a "C" rating in this pull request. CodeClimate isn't a required pass to pull in a pull request, so we can ignore the issue if we want.

templates/cat1/r3/_2.16.840.1.113883.10.20.24.3.105.cat1.erb
```diff
@@ -0,0 +1,68 @@
...
+    <code code="10183-2"
+          codeSystem="2.16.840.1.113883.6.1" 
+          codeSystemName="LOINC" 
+          displayName="Discharge medication"/> 
+    <statusCode code="active"/>
+    <effectiveTime>
+      <low <%= value_or_null_flavor(entry.start_time) %>/>
+    </effectiveTime>
+    <entryRelationship typeCode="SUBJ">
+      <substanceAdministration moodCode="EVN" classCode="SBADM" <%== negation_indicator(entry) %>>
```
dczulada: should the negation only be at the "act" level

templates/cat1/r3/_2.16.840.1.113883.10.20.24.3.105.cat1.erb
```diff
@@ -0,0 +1,68 @@
...
+            <id root="<%= UUID.generate %>"/>
+            <manufacturedMaterial>
+              <%== code_display(entry,'preferred_code_sets' => ['RxNorm'],'value_set_map' => filtered_vs_map, 'extra_content' => "sdtc:valueSet=\"#{value_set_oid}\"") %>
+            </manufacturedMaterial>
+            <manufacturerOrganization>
+              <name>Medication Factory Inc.</name>
+            </manufacturerOrganization>
+          </manufacturedProduct>
+        </consumable>
+        <%== render(:partial => 'reason', :locals => {:entry => entry, :reason_oids=>field_oids["REASON"]}) %>
```
dczulada: Should the reason be at the "act" level?

templates/cat1/r3/_2.16.840.1.113883.10.20.24.3.105.cat1.erb
mayerm94: The substanceAdministration 2.16.840.1.113883.10.20.24.3.41:2014-12-01 template may have a negationInd. Is there a constraint on sub-templates containing negationInd that I am not aware of?

templates/cat1/r3/_2.16.840.1.113883.10.20.24.3.105.cat1.erb
mayerm94: Reason is an optional field within the substanceAdministration template 2.16.840.1.113883.10.20.24.3.41:2014-12-01

templates/cat1/r3/_2.16.840.1.113883.10.20.24.3.105.cat1.erb
dczulada: This change would just make it consistent with the R3_1 template.  I can't find an explicit rule that states that the negation needs to be at the QDM data type level, but i feel like its more clear at that level.

2.16.840.1.113883.10.20.24.3.105 defines the template for QDM datatype: Medication, Discharge.  Therefore it makes more sense to me to represent a "Medication, Discharge not done" with the negationInd=true at the "Medication Discharge" level instead of nested in the Medication Active template.

dczulada (approved):
Changes look good

PR #452, opened 2016-11-29
BONNIE-207 Add cache reset to beginning of export.

alexanderelliott121:
Adds a cache clear to the scooped_view_helper.
Invokes scooped_view_helper cache clear at the beginning of export.

### DIFF
--- a/lib/health-data-standards/export/cat_1.rb
+++ b/lib/health-data-standards/export/cat_1.rb
@@ -18,6 +18,7 @@ module HealthDataStandards
 
       #default qrda_version should default to latest version
       def export(patient, measures, start_date, end_date, header=nil, qrda_version='r3_1', cms_compatibility=false)
+        #this will ensure that any value_sets that have been added to a bundle since the last qrda export will be picked up.
         clear_vs_map(patient.bundle_id)
         
         qrda_version = qrda_version.nil? ? @@most_recent_qrda_version : qrda_version
--- a/lib/health-data-standards/export/cat_1.rb
+++ b/lib/health-data-standards/export/cat_1.rb
@@ -1,6 +1,7 @@
 module HealthDataStandards
   module Export
     class Cat1
+      include HealthDataStandards::Export::Helper::ScoopedViewHelper
       
       def initialize(qrda_version = nil)
         template_helper = HealthDataStandards::Export::TemplateHelper.new('cat1', 'cat1', nil, qrda_version)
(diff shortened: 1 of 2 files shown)


lib/health-data-standards/export/helper/scooped_view_helper.rb
```diff
@@ -6,6 +6,10 @@ module ScoopedViewHelper
         include HealthDataStandards::SVS
         VS_MAP = {}
 
+        def clear_vs_map(bundle_id=nil)
```
edeyoung, line 9: add comment explaining what this should be used for (clearing cache before each export)

lib/health-data-standards/export/cat_1.rb
```diff
@@ -17,7 +18,8 @@ def initialize(qrda_version = nil)
 
       #default qrda_version should default to latest version
       def export(patient, measures, start_date, end_date, header=nil, qrda_version='r3_1', cms_compatibility=false)
-
+        clear_vs_map(patient.bundle_id)
```
edeyoung, line 22: add comment explaining why the cache is being cleared.

lib/health-data-standards/export/helper/scooped_view_helper.rb
edeyoung, line 9: i don't think `bundle_id` should be optional here. What do you think? Or if it is `nil`, should we be grabbing `latest_bundle_id` as well, like is done below? or is that too risky?

edeyoung (changes requested):

lib/health-data-standards/export/helper/scooped_view_helper.rb
alexanderelliott121: It makes sense to use the latest bundle id, since that is the cache that will be accessed.

okeefm: Looks like @edeyoung's comment here was addressed (by grabbing `latest_bundle_id`) in the commit below, but the GitHub system didn't pick it up.

okeefm (approved):
Looks good to me.

edeyoung (approved):

edeyoung:
@okeefm it looks like these changes are against bonnie_master_2016 and should actually be applied to the master branch.

okeefm:
Cherry-picked onto `master` in PR #457.

PR #453, opened 2016-11-30
oid_for_code now returns vs_oid even if the code is the 0th element in the oid_map

mayerm94:
Change condition to check if index returns nil so that it does not evalute to false when code is at index 0

### DIFF
--- a/lib/health-data-standards/export/helper/cat1_view_helper.rb
+++ b/lib/health-data-standards/export/helper/cat1_view_helper.rb
@@ -52,7 +52,7 @@ module HealthDataStandards
           valueset_oids.each do |vs_oid|
             oid_list = (vs_map[vs_oid] || [])
             oid_map = Hash[oid_list.collect{|x| [x["set"],x["values"]]}]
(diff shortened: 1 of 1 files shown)


lib/health-data-standards/export/helper/cat1_view_helper.rb
```diff
@@ -52,7 +52,7 @@ def oid_for_code(codedValue, valueset_oids,  bundle_id = nil)
           valueset_oids.each do |vs_oid|
             oid_list = (vs_map[vs_oid] || [])
             oid_map = Hash[oid_list.collect{|x| [x["set"],x["values"]]}]
-            if (oid_map[code_system] || []).index code
+            if ((oid_map[code_system] || []).index code).nil?
```
okeefm: Rather than relying on a comparison to the return value of `.index`, which is what caused this bug, we should use `Array#include?`, by doing something like `if (oid_map[code_system] || []).include? code`

okeefm (dismissed):

edeyoung (approved):

tstrass (approved):

PR #460, opened 2016-12-20
time 'field names' rely on an hqmf entry to have a template_id.  Some…

dczulada:
…times the template_id is in the outboundRelationships entry


dczulada:
screen shot 2016-12-20 at 10 53 11 am

screen shot 2016-12-20 at 10 53 44 am

okeefm (changes requested):
My only comment is that `outboundRelationship` should be `outbound_relationship`, because Ruby in general prefers snake case for variables rather than camel case.

Also this should be rebased.

PR #461, opened 2016-12-20
Null flavor negation

dczulada:


lib/health-data-standards/import/cda/section_importer.rb
```diff
@@ -189,6 +192,28 @@ def extract_reason_or_negation(parent_element, entry)
           elsif negation_indicator
             entry.negation_ind = negation_indicator.eql?('true')
           end
+          extract_negated_code(coded_parent_element, entry)
+        end
+
+        def extract_negated_code(coded_parent_element, entry)
+          code_elements = coded_parent_element.xpath(@code_xpath)
+          code_elements.each do |code_element|
+            if code_element['nullFlavor'] == 'NA' && code_element['sdtc:valueSet']
```
okeefm, line 201: We should think about adding a way to tell that we've added in a code here (such as a flag on the `Entry`) so we can troubleshoot this.

okeefm (changes requested):

Issue #333, opened 2015-11-26
Patient Importer - Comm from Provider to Provider - Wrong ID


grexican:
In `patient_importer.rb`:
`generate_importer(Cat1::ProcedureImporter, nil, '2.16.840.1.113883.3.560.1.129'), #comm from provider to provider`

If I look at the spec, the code should be `...1.29` not `...`1.129`. There's an extra 1 in there.`1.129`is the`negation: true`version of`communication_from_provider_to_provider`according to the`hqmf_template_oid_map`.

I think anyway :) If I'm wrong, please let me know why. FWIW, I had to edit my `records` documents in my DB to be `1.29` instead of `1.129` to pass `Measure 0089 QRDA Cat I Test` in Cypress.

grexican:
The following monkeypatch, under the assumption it should be 1.29 and not 1.129, makes measure 0089 passable in cypress.

```
module HealthDataStandards
  module Import
    module Cat1

      # This class is the central location for taking a QRDA Cat 1 XML document and converting it
      # into the processed form we store in MongoDB. The class does this by running each measure
      # independently on the XML document
      #
      # This class is a Singleton. It should be accessed by calling PatientImporter.instance
      class PatientImporter
        alias_method :original_initialize, :initialize

        def initialize
          original_initialize

          # MONKEY PATCH: USING 1.129 (negative) instead of 1.29 (positive) for P2P Communication
          begin
            index = @section_importers[:procedures].find_index { |p| p.hqmf_oid == '2.16.840.1.113883.3.560.1.129'}

            if index.present?
              ptpcomm = @section_importers[:procedures][index]
              ptpcomm.hqmf_oid = '2.16.840.1.113883.3.560.1.29'
              @section_importers[:procedures][index] = ptpcomm

              puts "monkey patched initialize: #{@section_importers[:procedures][index].to_yaml}"
            else
              puts "MONKEY PATCH FAILED!"
            end

          rescue
            puts "Error #{$!}"
          end
        end

      end
    end
  end
end
```

grexican:
Actually the issue, I think, is much bigger than this. As I said above, it's not just the OID, but it's the _NEGATIVE_ OID. Well that negative (_NOT DONE_) version needs to be accounted for to. And from what I see, it's not accounted for anywhere in the importer.

[...]

PR #463, opened 2016-12-22
This stores/sets the negated/non-negated hqmf_oid by checking negation_ind=true while importing

nadeemnazeer3:
This stores/sets the negated/non-negated hqmf_oid by checking negation_ind=true(This caused errors while Cypress testing)
-  Fixes issues related to this issue: https://github.com/projectcypress/health-data-standards/issues/333 
- This will thus handle proper matching while looking for entries and later the exporting of proper entries/templates
- It supposes We keep non-negated hqmf_oid as default in patient importer for different sections (in lib/health-data-standards/import/cat1/patient_importer.rb)
- This changes the last part of hqmf_oid to negate it by adding 100 to it
- Example: 2.16.840.1.113883.3.560.1.6 will be changed to its negation 2.16.840.1.113883.3.560.1.106

### DIFF
--- a/lib/health-data-standards/import/cat1/entry_package.rb
+++ b/lib/health-data-standards/import/cat1/entry_package.rb
@@ -13,12 +13,24 @@ module HealthDataStandards
 
         def package_entries (doc, nrh)
           entries = self.importer_type.create_entries(doc, nrh)
-          entries.each do |entry| 
-            entry.oid = self.hqmf_oid
+          entries.each do |entry|
+            # This stores/sets the negated/non-negated hqmf_oid by checking negation_ind=true(This caused errors while Cypress testing)
+            # This will thus handle proper matching while looking for entries and later the exporting of proper entries/templates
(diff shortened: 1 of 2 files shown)


nadeemnazeer3:
This allows correctly setting proper hqmf_id for negated entries while importing. 
There was no distinction made while importing negated/non-negated entries e.g: In patient importer for **"communication from provider to provider not done"** hqmf_id was set that is : 2.16.840.1.113883.3.560.1.129 but nothing changed or considered while importing **"communication from provider to provider done"** for which hqmf_id should have been set to: 2.16.840.1.113883.3.560.1.29.
likewise for other sections with negated/non-negated hqmf_ids.

nadeemnazeer3:
This PR fixes the issue like this: https://github.com/projectcypress/health-data-standards/issues/333 @okeefm @dczulada just tagging as this seems crtitical and while using the HDS may cause Cypress testing fails for many, (Just i felt Sharing is Caring :) and contributing back)

PR #465, opened 2016-12-27
QRDA cat III R2 added. Qrda Cat III versions moved into folders like …

dczulada:
…cat I


edeyoung (approved):
I have not run this and did not look at the r1_1 templates. However, the logic changes made and the changes to the r1 templates look sound to me.

I recommend another review by a Cypress team member who is more familiar with the QRDA Cat 3 templates.

okeefm:
Like @edeyoung said, all the code (not QRDA) changes look good to me. I'm going to take a closer look this weekend at the QRDA R1/R1_1/R2 templates and make sure they're consistent with the IG.

templates/cat3/r2/_measure_data.cat3.erb
```diff
@@ -0,0 +1,136 @@
...
+<observation classCode="OBS" moodCode="EVN">
+  <!-- Measure Data template -->
+  <templateId root="2.16.840.1.113883.10.20.27.3.5" extension="2016-09-01"/>
+  <code code="ASSERTION" 
+        codeSystem="2.16.840.1.113883.5.4" 
+        displayName="Assertion" 
+        codeSystemName="ActCode"/>
+  <statusCode code="completed"/>
+  <value xsi:type="CD" code="<%= population.type == 'IPP' ? 'IPOP' : population.type %>"
+         codeSystem="2.16.840.1.113883.5.4"
```
okeefm, line 12: Do we care about including the population display name here? Probably not super important.

dczulada: have we ever?  I don't believe there is huge value added there

okeefm: I don't think we have, so that's fine.

PR #466, opened 2016-12-29
HQMF parser now preserves the full HQMF eMeasure version number

edeyoung:
the HQMF file now have finer granularity for version numbers (e.g. 5.1.000). With the Measure History View in Bonnie, this will be helpful for the users to differentiate between uploads of the same version of the measure. To capture this, the version should be stored as a string rather than an integer.


dczulada (approved):
Looks good

okeefm (approved):
Looks good to me.

okeefm:
Needs to be rebased from master, then should be good to merge.

PR #467, opened 2017-01-04
Change condition that checks if code_system exists

mayerm94:
Change condition that checks if code_system exists to return true if code_system is in the 0th location

Previous logic used just .include which returns index of code_system in map, if the index is 0 the logic evaluates to false. I changed the logic to check if the .include call returns nil instead

### DIFF
--- a/lib/health-data-standards/export/helper/cat1_view_helper.rb
+++ b/lib/health-data-standards/export/helper/cat1_view_helper.rb
@@ -52,7 +52,7 @@ module HealthDataStandards
           valueset_oids.each do |vs_oid|
             oid_list = (vs_map[vs_oid] || [])
             oid_map = Hash[oid_list.collect{|x| [x["set"],x["values"]]}]
-            if (oid_map[code_system] || []).include? code
+            if ((oid_map[code_system] || []).index code).nil?
               return vs_oid
             end
           end


lib/health-data-standards/export/helper/cat1_view_helper.rb
```diff
@@ -52,7 +52,7 @@ def oid_for_code(codedValue, valueset_oids,  bundle_id = nil)
           valueset_oids.each do |vs_oid|
             oid_list = (vs_map[vs_oid] || [])
             oid_map = Hash[oid_list.collect{|x| [x["set"],x["values"]]}]
-            if (oid_map[code_system] || []).include? code
```
okeefm, line 55: There's no need to completely change this logic to do a `nil` comparison; you can just invert the value of the check here (or use `unless` instead of `if`).

okeefm: Also, we should think about improving the testing around this function, given that this is the second pull request changing this line.

okeefm: Also also I apologize that I accidentally inverted the check value of this `if` statement in #453, which necessitated this PR.

PR #468, opened 2017-01-04
reverses changes in BONNIE-207

dczulada:


okeefm (approved):
Only for Cypress's `mongoid5` branch, so looks good to me.

PR #474, opened 2017-02-08
only import value if it isn't a nullFlavor

dczulada:


okeefm (approved):
Seems reasonable to me.

edeyoung (commented):
Not relevant to Bonnie so not reviewing and I don't think the Bonnie team needs to review.

PR #475, opened 2017-02-13
several displayed descriptions for data criteria needed to change to …

edeyoung:
…conform with QDM 5.0 changes. However, the codes themselves seemed like they should stay the same for backwards compatibility. this led to discrepencies in the testing. these are addressed in this pull request.

### DIFF
--- a/test/unit/hqmf/model/data_criteria_test.rb
+++ b/test/unit/hqmf/model/data_criteria_test.rb
@@ -16,11 +16,12 @@ module HQMFModel
         mismatches << field_title unless field == title
       end
 
-      # ignore custom fulfills and previous ordinal field values
-      mismatches -= ["FLFS - FULFILLS", "ORDINAL - ORDINALITY"]
-
-      # ignore changes needed when converting to qdm 5.0
-      mismatches -= [
+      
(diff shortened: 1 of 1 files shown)


test/unit/hqmf/model/data_criteria_test.rb
```diff
@@ -19,6 +19,16 @@ def test_fields_definitions
       # ignore custom fulfills and previous ordinal field values
       mismatches -= ["FLFS - FULFILLS", "ORDINAL - ORDINALITY"]
 
+      # ignore changes needed when converting to qdm 5.0
```
c-monkey: Should we complain if these mismatches are not found? I think that could only happen if we update data_criteria.rb to remove one of these fields.

c-monkey (approved):
Test passes.

test/unit/hqmf/model/data_criteria_test.rb
edeyoung: interesting idea! check out my change.

jbradl11 (approved):

c-monkey:
I like the changed test.

PR #477, opened 2017-03-01
Correct Procedure Order TemplateId

dczulada:
The current templateId extension in hds is 2014-12-01.  It should be 2016-02-01 instead.


okeefm (approved):
Verified in the 3.1 IG.

edeyoung (approved):

PR #479, opened 2017-03-06
Now capturing missing referenced data criteria ids in the HQMF parser

holmesie:
Referenced data criteria were not being included in the reference_ids array during HQMF parsing when those data criteria were in fact referenced by observation criteria. This caused the unnecessary removal of data criteria during the filtering phase. When extracting observation criteria, the parser now records any children criteria as referenced, so they will not be removed later in the parsing process.


hossenlopp (approved):

dczulada (approved):
Looks good

edeyoung (approved):

PR #480, opened 2017-03-07
Round expected performance rate in error message

dczulada:


okeefm (approved):
Looks good.

edeyoung (commented):
Not relevant to Bonnie code, so I will not submit a review and I don't think the Bonnie team needs to review.

PR #481, opened 2017-03-21
Change Diagnostic Study, Recommended HQMFR1 OID to be correct

okeefm:
`Diagnostic Study, Recommended` used to have the same HQMF OID in the map as `Diagnostic Study, Ordered`. This didn't cause issues in HDS, because we have no `Diagnostic Study, Recommended` elements, and `Diagnostic Study, Ordered` comes first in HDS. However, `Go` doesn't respect orderings like this (for precisely this reason), so we were sometimes getting the `Diagnostic Study, Recommended` QRDA OID when we wanted the `Diagnostic Study, Ordered` one. This change changes the HQMF OID to be the correct one, just in case HDS has a `Diagnostic Study, Recommended` entry in the future.

### DIFF
--- a/lib/health-data-standards/export/qrda/hqmf-qrda-oids.json
+++ b/lib/health-data-standards/export/qrda/hqmf-qrda-oids.json
@@ -115,7 +115,7 @@
   },
   {
     "hqmf_name": "Diagnostic Study, Recommended",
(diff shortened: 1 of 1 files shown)


edeyoung (approved):

dczulada (approved):
Looks good

PR #482, opened 2017-03-24
Capture the observations from CQL

mulcahyk:
Update the CQL parsing to pull out the measure observations on continuous variable measures.  This fixes the backend.


lib/hqmf-parser/cql/document_helpers/doc_population_helper.rb
```diff
@@ -0,0 +1,67 @@
...
+      @reference_ids = reference_ids
+    end
+
+    def extract_populations
+      @populations_cql_map = extract_populations_cql_map
+      @observations = extract_observations
+      [@populations_cql_map, @observations]
+    end
+
+    # Extracts the measure observations, will return true if one exists
```
holmesie: I just noticed this comment is no longer true, since it will always return at least an empty array. Might want to just make it say: "Extracts potential measure observations from the CQL based HQMF".

lib/hqmf-parser/cql/document_helpers/doc_population_helper.rb
```diff
@@ -0,0 +1,67 @@
...
+    def extract_observations
+      observations = []
+
+      # look for observation data in separate section but create a population for it if it exists
```
holmesie: Same as above, this comment is sort of outdated now. Might want to say something like: "Look for observations in the measureObservationSection of the CQL based HQMF document, and if they exist extract the name of the CQL statement that calculates the observation.".

edeyoung (approved):

jbradl11 (approved):

PR #483, opened 2017-03-29
Qdm 5.01 cql

mayerm94:
These include the addition of the components attribute added in the QDM 5.01 document.


lib/health-data-standards/models/procedure.rb
```diff
@@ -17,7 +17,9 @@ class Procedure < Entry
 
   # QDM 5.0 addition to Diagnostic Study, Performed
   field :result_date_time,    type: Integer
-
+  # Component attribute is for Diagnostic, Performed which is classified as a procedure
+  field :components,           type: Hash
```
edeyoung: nit pick, but watch spacing for things like `type: Hash` where they are aligned. Similar for other files.

lib/hqmf-model/data_criteria.rb
```diff
@@ -34,6 +34,8 @@ class DataCriteria
               'ANATOMICAL_LOCATION_SITE' => {title:'Anatomical Location Site', coded_entry_method: :anatomical_location,  field_type: :value},
               'ANATOMICAL_STRUCTURE' => {title:'Anatomical Structure', coded_entry_method: :anatomical_structure, code: '91723000', code_system:'2.16.840.1.113883.6.96', template_id: '2.16.840.1.113883.3.560.1.1000.2', field_type: :value},
               'CAUSE' => {title:'Cause', coded_entry_method: :cause_of_death, code: '42752001', code_system:'2.16.840.1.113883.6.96', template_id: '2.16.840.1.113883.3.560.1.1017.2', field_type: :value},
+              # TODO: Determine actual code and code_system for component attribute
+              'COMPONENT' => {title: 'Component', coded_entry_method: :components, code: '1234567', code_system: '2.16.840.1.113883.6.96', field_type: :value},
```
edeyoung: rather than use fake codes, remove the `code` and `code_system` fields. When a fake code is necessary, use something very obvious. For example, make the `code_system` here by `1.11.111.111...`.

See `RESULT_DATETIME` as an example.

lib/hqmf-model/data_criteria.rb
```diff
@@ -98,6 +100,8 @@ class DataCriteria
     # maps attribute codes to the attribute keys
     VALUE_FIELDS = {'399423000' => 'ADMISSION_DATETIME',
                     '42752001' => 'CAUSE',
+                    # TODO: Fill in actual code for component
+                    '1234567'   => 'COMPONENTS',
```
edeyoung: You can remove this given the suggestion to remove the code from above. `VALUE_FIELDS` not necessary for CQL parsing.

lib/hqmf-model/data_criteria.rb
```diff
@@ -441,6 +445,10 @@ def self.convert_value(json)
         when 'ACT'
           # Currentlty forcing this as the SimpleXML reresentation contains a fulfills for these types
           value = HQMF::TypedReference.new(json["reference"], 'FLFS', '')
+        when 'CMP'
+          value = HQMF::Component.from_json(json)
+        when 'COL'
+          value = HQMF::GenericCollection.from_json(json)
```
edeyoung: Consider renaming from `GenericCollection` to `Collection`

lib/hqmf-model/types.rb
```diff
@@ -155,7 +155,6 @@ def type
   class Coded
     include HQMF::Conversion::Utilities
     attr_reader :type, :system, :code, :code_list_id, :title, :null_flavor, :original_text
-
```
edeyoung, line 158: be careful of changes that are just whitespace changes. Try to catch these when doing diffs pre-commit.

lib/hqmf-model/types.rb
```diff
@@ -428,4 +425,65 @@ def ==(other)
...
+      @type = type || 'COL'
+      @values = values || []
+    end
+
+    def self.from_json(json)
+      json = json.with_indifferent_access
+      values = []
+      type = json['type']
+      json['values'].each { |value| values.push(HQMF::DataCriteria.convert_value(value))}
+      HQMF::GenericCollection.new(type,values)
```
edeyoung: watch spacing: `type,values` -> `type, values`

lib/hqmf-model/types.rb
```diff
@@ -428,4 +425,65 @@ def ==(other)
...
+  end
+  
+  class Component
+    include HQMF::Conversion::Utilities
+    # From QDM 5.02 "A Component attribute includes a code and result"
+    # "for laboratorytest performed includes optional reference range high and reference range low"
+    attr_accessor :type, :code, :result, :range
+    
+    def initialize(code, value, unit, code_list_id, title, low = nil, high = nil)
+      @type = 'CMP'
```
edeyoung: Should this be `@type = type || 'COM'` like the others are?

lib/hqmf-model/types.rb
```diff
@@ -428,4 +425,65 @@ def ==(other)
...
+    def self.from_json(json)
+      code = json['code'] if json['code']
+      HQMF::Component.new(code, json['value'], json['unit'], json['code_list_id'], json['title'])
+    end
+
+    def to_json
+      hash = build_hash(self, [:type])
+      hash['code'] = @code.to_json
+      hash['result'] = @result.to_json
+      #hash[:range] = @range.to_json if @range
```
edeyoung: add TODO here

lib/hqmf-parser/cql/document.rb
```diff
@@ -40,6 +40,7 @@ def extract_criteria
         sdc = dc.clone # Clone data criteria
         sdc.id += '_source' # Make it a source
 
+        # REVIEW: For HQMF + CQL, do we need both DC and SDC?
```
edeyoung: remove comment this.

for now, i think we do because of how the bonnie interface uses both DC and SDC. If we were to rework that, maybe not.
@holmesie what are your thoughts?

lib/hqmf-parser/cql/document.rb
```diff
@@ -63,13 +64,8 @@ def extract_populations_cql_map
...
             cql_statement = criteria_def.at_xpath("*/*/cda:id", HQMF2::Document::NAMESPACES).attribute('extension').to_s.match(/"([^"]*)"/)
-            if @populations_cql_map[criteria_id].nil?
-              @populations_cql_map[criteria_id] = []
-            end
-            cql_statement = cql_statement.to_s.delete('\\"')
-            unless @populations_cql_map[criteria_id].include? cql_statement
-              @populations_cql_map[criteria_id].push(cql_statement)
-            end
+            @populations_cql_map[criteria_id] = cql_statement.to_s
+            @populations_cql_map[criteria_id] = cql_statement.to_s.delete('\\"')
```
edeyoung: was this necessary for your changes or a simplification that you noticed could be implemented? if a simplification, please revert and include in another PR so we can test any effects of this separately. if necessary, please explain. the code looks like it removes some checks and default behavior that may be important.

lib/hqmf-parser/cql/population_helper.rb
```diff
@@ -0,0 +1,12 @@
+module HQMF2CQL
```
edeyoung: is this class used anywhere or necessary? It looks like it might be left-over from a previous idea.

test/unit/hqmf/cql/hqmf_cql_test.rb
```diff
@@ -17,7 +17,7 @@ class HQMF2CQLTest < Minitest::Test
 
   Dir.glob(measure_files).each do |measure_filename|
     measure_name = File.basename(measure_filename, '.xml')
-    if ['2'].index(measure_name)
+    if ["1"].index(measure_name)
```
edeyoung, line 20: why is this updated to `'1'`?

edeyoung (changes requested):

test/unit/hqmf/cql/hqmf_cql_test.rb
mayerm94: Not sure, this was pulled in from the rebase in commit https://github.com/projectcypress/health-data-standards/pull/483/commits/bb5d9b458c718301236c696a48483c70d746773c

mayerm94: It is worth noting that there is no difference in test result when either 1 or 2 is used

lib/hqmf-parser/cql/population_helper.rb
mayerm94: I believe this class is still used, the changes to incorporate stratification parsing (which was reverted) was originally made in this file in the extract_populations_cql_map. I believe that particular function has since been moved to document_helpers/doc_population_helper.rb

edeyoung (approved):

edeyoung:
still need to test end-to-end with the other changes made to support QDM 5.01

edeyoung (changes requested):
the reference to population_helper needs to be removed from health-data-standards/lib/hqmf-parser.

edeyoung:
There were issues with the commit stack for this request. this got addressed in this PR: https://github.com/projectcypress/health-data-standards/pull/490

PR #484, opened 2017-03-31
Bug fix of Bonnie CQL support for stratifications

c-monkey:
missing stratifications from HQMF parsing block for CQL


jbradl11 (approved):

edeyoung (approved):

edeyoung:
This PR needs to be reverted or wait until all stratification PRs are incorporated.
This adds a stratification to measures when it isn't necessary because the HQMF always seems to have a stratification (related to race/payer/etc.) (Matt took a look and saw this).
This makes everything fail because the stratification always calculates to zero, so the code decides no populations should calculate (which makes sense).

PR #485, opened 2017-04-05
Add bundler-audit and update gems to comply

tstrass:
bundler-audit identifies gem versions with known vulnerabilities.

This adds bundler-audit to the project and brings HDS into compliance.
Previously out of compliance: nokogiri, activemodel, rubyzip

Note that the [`zip-zip`](https://github.com/orien/zip-zip) gem was added because `rubyzip` needed to be updated to a version that changed their API, and `zip-zip` is apparently a way to make it backwards compatible.


tstrass:
I did not add `bundler-audit` to Travis, but I can add that in if reviewers want it

dczulada (approved):
Looks good

edeyoung (commented):
Someone from the Bonnie team needs to test this to make sure it doesn't mess things up for us. I'll mention it at standup today and then assign someone (both at standup and on the ticket).

edeyoung:
Initial pass shows there are some conflicts with bonnie that are non-trivial to address. @hossenlopp is looking into it.

okeefm (changes requested):
Regardless of the Bonnie reviews, I would suggest that if we're using Bundler-audit, it should be in Travis.

okeefm (approved):
Travis change looks good to me.

PR #487, opened 2017-04-07
Fix support for CV Measures

mulcahyk:
Updated the way that the observations in Continuous Variable measures are stored.  Based on an deeper understanding of the specs CV measures will have a function (defined in the CQL) and a parameter (defined in the HQMF) that is used for the aggregation.  This pulls these two pieces of information out of the HQMF.


holmesie (approved):
Looks good.

jbradl11 (approved):

PR #492, opened 2017-04-20
Gem security updates

tstrass:
This contains commits cherry-picked from another branch which is currently in an open PR to master, which is stalled in review: #485

We are doing this because HDS failing bundler-audit is currently blocking all PRs on Cypress. If any further changes come out of #485, they will be subsequently incorporated into `mongoid5` as well.


okeefm (approved):
Looks good to me.

PR #495, opened 2017-05-08
BONNIE-589

mulcahyk:
Make string compare case-insensitive
When checking for negations the code was looking for 'true'.  It appears that the MAT has started to create HQMF files with the actionNegationInd with 'TRUE'.


c-monkey (approved):

holmesie (approved):

dczulada (approved):
looks good

PR #500, opened 2017-06-06
Mongoid5 qrda4

dczulada:


okeefm (dismissed):
I'm able to reproduce the Travis failures locally, so something seems actually wrong here.

okeefm (approved):
Looks good to me; incorporates the `<statusCode>` change.

PR #502, opened 2017-06-06
Extract main library name and return with model

jbradl11:
Extracts the name of the measure and returns the value with the model. In order to test checkout the bonnie_bundler branch "multi_library_support".  Any CQL measure should work for testing.

### DIFF
--- a/lib/hqmf-parser/cql/document_helpers/doc_population_helper.rb
+++ b/lib/hqmf-parser/cql/document_helpers/doc_population_helper.rb
@@ -68,5 +68,17 @@ module HQMF2CQL
       end
       populations_cql_map
     end
+
+    # Extracts the name of the main cql library from the Population Criteria Section
+    def extract_main_library
+      population_criteria_sections = @doc.xpath("//cda:populationCriteriaSection/cda:component[@typeCode='COMP']", HQMF2::Document::NAMESPACES)
+      criteria_section = population_criteria_sections.at_xpath("cda:initialPopulationCriteria", HQMF2::Document::NAMESPACES)
+      if criteria_section
+        cql_main_library_name = criteria_section.at_xpath("*/*/cda:id", HQMF2::Document::NAMESPACES).attribute('extension').to_s.match(/[^.]*/).to_s
+      else
+        nil
+      end
+    end
+
   end
 end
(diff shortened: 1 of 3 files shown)


mulcahyk:
Have we reached out to the MAT to find out how they are going to be doing the export?  Is there something in the HQMF spec for this?

lib/hqmf-parser/cql/document_helpers/doc_population_helper.rb
```diff
@@ -68,5 +68,17 @@ def extract_populations_cql_map
       end
       populations_cql_map
     end
+
+    # Extracts the name of the main cql library from the Population Criteria Section
+    def extract_main_library
+      population_criteria_sections = @doc.xpath("//cda:populationCriteriaSection/cda:component[@typeCode='COMP']", HQMF2::Document::NAMESPACES)
+      criteria_section = population_criteria_sections.at_xpath("cda:initialPopulationCriteria", HQMF2::Document::NAMESPACES)
+      if criteria_section
+        cql_main_library_name = criteria_section.at_xpath("*/*/cda:id", HQMF2::Document::NAMESPACES).attribute('extension').to_s.match(/[^.]*/).to_s
```
mulcahyk: Can you explain why this regex pattern?

jbradl11:
@mulcahyk  We have seen the standards and we have seen the export. The HQMF spec in the population criteria section contains the name of the measure file.

lib/hqmf-parser/cql/document_helpers/doc_population_helper.rb
```diff
@@ -68,5 +68,19 @@ def extract_populations_cql_map
...
+    # Extracts the name of the main cql library from the Population Criteria Section
+    def extract_main_library
+      population_criteria_sections = @doc.xpath("//cda:populationCriteriaSection/cda:component[@typeCode='COMP']", HQMF2::Document::NAMESPACES)
+      criteria_section = population_criteria_sections.at_xpath("cda:initialPopulationCriteria", HQMF2::Document::NAMESPACES)
+      if criteria_section
+        # Example: the full name for the population criteria section is "BonnieNesting01.\"Initial Population\""
+        # The regex returns everything before the "."  (BonnieNesting01), which is the file name of the cql measure
+        cql_main_library_name = criteria_section.at_xpath("*/*/cda:id", HQMF2::Document::NAMESPACES).attribute('extension').to_s.match(/[^.]*/).to_s
+      else
+        nil
```
mulcahyk, line 81: We should consider having some way to let the end user know that the primary library could not be found.

jbradl11: I believe this would be done in Bonnie Bundler. We try to grab the value, if the value is nil, we bubble up an exception.

holmesie (approved):

mulcahyk (approved):

PR #505, opened 2017-06-27
Add support for CAT I R4 in QDM Template Validator

dczulada:


dczulada:
@edeyoung this only impacts a validator, should only need a 'quick' review from bonnie

test/unit/validate/qrda_qdm_template_validator_test.rb
```diff
@@ -5,12 +5,12 @@ class QrdaQdmTemplateValidatorTest < ActiveSupport::TestCase
   setup do
     @validator_r3 = HealthDataStandards::Validate::QrdaQdmTemplateValidator.new('r3')
     @validator_r3_1 = HealthDataStandards::Validate::QrdaQdmTemplateValidator.new('r3_1')
-    @validator_r4 = HealthDataStandards::Validate::QrdaQdmTemplateValidator.new('r4')
+    @validator_r1 = HealthDataStandards::Validate::QrdaQdmTemplateValidator.new('r1')
```
edeyoung, line 8: why is the name changed? is `r1` ever used?

dczulada: The test is "Should not produce errors if validator does not support version specified"

I changed it to r1 specifically because it will never be supported.

edeyoung (approved):
Not tested by me, but looks good otherwise. I assume O'Keefe is testing.

okeefm (approved):
Tested and validated the `QRDA_Cat_1_R4_QDM_OIDS` list.

PR #506, opened 2017-06-27
Added logic to further support observations in HQMF+CQL

holmesie:
This is the first of three pull requests; the other two are:
2: https://github.com/projecttacoma/bonnie_bundler/pull/87
3: https://github.com/projecttacoma/bonnie/pull/765

Measure to test with: 
[CMS55_v5_1_Artifacts.10.zip](https://github.com/projectcypress/health-data-standards/files/1106977/CMS55_v5_1_Artifacts.10.zip)

### DIFF
--- a/lib/hqmf-parser/2.0/population_criteria.rb
+++ b/lib/hqmf-parser/2.0/population_criteria.rb
@@ -34,7 +34,7 @@ module HQMF2
       else
         # Extract the data criteria this population references
         dc = handle_observation_criteria
-        @preconditions = [Precondition.new(id_generator.next_id, nil, nil, false, HQMF2::Reference.new(dc.id))]
+        @preconditions = [Precondition.new(id_generator.next_id, nil, nil, false, HQMF2::Reference.new(dc.id))] if dc
       end
     end
 
--- a/lib/hqmf-parser/cql/document.rb
+++ b/lib/hqmf-parser/cql/document.rb
@@ -20,9 +20,9 @@ module HQMF2CQL
 
     # Generates this classes hqmf-model equivalent.
     def to_model
-      dcs = all_data_criteria.collect(&:to_model)
-      sdc = source_data_criteria.collect(&:to_model)
-      pcs = all_population_criteria.collect(&:to_model)
+      dcs = all_data_criteria.compact.collect(&:to_model)
+      sdc = source_data_criteria.compact.collect(&:to_model)
+      pcs = all_population_criteria.compact.collect(&:to_model)
       HQMF::Document.new(@id, @id, @hqmf_set_id, @hqmf_version_number, @cms_id,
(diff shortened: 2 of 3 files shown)


lib/hqmf-parser/cql/document.rb
```diff
@@ -20,9 +20,9 @@ def initialize(hqmf_contents, use_default_measure_period = true)
 
     # Generates this classes hqmf-model equivalent.
     def to_model
-      dcs = all_data_criteria.collect(&:to_model)
-      sdc = source_data_criteria.collect(&:to_model)
-      pcs = all_population_criteria.collect(&:to_model)
+      dcs = all_data_criteria.compact.collect(&:to_model)
+      sdc = source_data_criteria.compact.collect(&:to_model)
+      pcs = all_population_criteria.compact.collect(&:to_model)
```
mulcahyk, line 25: Would it break a lot to change the variable names to something more descriptive?

lib/hqmf-parser/2.0/population_criteria.rb
```diff
@@ -34,7 +34,7 @@ def handle_type(id_generator)
       else
         # Extract the data criteria this population references
         dc = handle_observation_criteria
-        @preconditions = [Precondition.new(id_generator.next_id, nil, nil, false, HQMF2::Reference.new(dc.id))]
+        @preconditions = [Precondition.new(id_generator.next_id, nil, nil, false, HQMF2::Reference.new(dc.id))] if dc
```
mulcahyk, line 37: Can you add an explanation of why you are calling the method with these parameters?
I'm also not sure I understand the relationship of preconditions to observations.

holmesie: Note that this is a change to the HQMF w/ QDM datatypes AND QDM logic parser (the older one, i.e. 2.0), and the change is specifically nothing more than a sanity check.

Observations at this point had data criteria preconditions. Since this is no longer the case with the CQL+HQMF parser, and since our goal was to reuse as much of the older work as possible, I've simply added a check here in order to not create a precondition if there was originally no DC (as is the case w/ CQL).

lib/hqmf-parser/cql/document.rb
holmesie: These variable names originally come from the 2.0 parser. I am definitely open to renaming them in this version of the parser, but I would note that the description of what they contain are exactly what is to the right of each of the `=`. My change here is another sanity check, i.e. make sure no `nil` or duplicated criteria are present before calling the to_model method on each index in these arrays.

edeyoung (approved):
reviewed over phone with adam

mayerm94 (approved):
Tested locally with the supplied measure.

mulcahyk (approved):
I understand the part about preconditions now.

PR #509, opened 2017-06-30
Import CMS codes in the supplemental data extractor

ljades:


lib/health-data-standards/validate/reported_result_extractor.rb
```diff
@@ -152,7 +152,12 @@ def extract_supplemental_data(cv)
             value = node.at_xpath('cda:value')
             count = get_aggregate_count(node)
             if value.at_xpath("./@nullFlavor")
-             key_hash["UNK"] = count
+             if supp == "PAYER" && value['xsi:type'] == 'CD' && value['nullFlavor'] == 'OTH' && value.at_xpath("cda:translation") &&
+                value.at_xpath("cda:translation")['code']
```
okeefm: Is there a reason just this one conditional from the `if` statement is on the next line? I think it makes the conditional harder to read.

okeefm (approved):
Looks good to me. Just need a review from the Bonnie team.

edeyoung (approved):
Untested but looks good and like it will have minimal impact on Bonnie.

PR #513, opened 2017-07-12
Add a CCN getter/setter to Provider

okeefm:
CCN should be stored with the Provider so we can output it into the Cat I without hardcoding it into the template. This adds code to do that.


dczulada (approved):
Looks good

okeefm:
Note: this was closed and reopened to get Travis to run.

PR #514, opened 2017-07-14
Support for MAT 5.3 exports in the CQL+HQMF parser

holmesie:
Includes support for CQL based HQMF exports from the MAT version 5.3 - while keeping (temporary) support for older measures. The big difference is the template OIDs changed between versions.

MAT 5.3 measures to experiment with:
[CMS158_v5_3_Artifacts.zip](https://github.com/projectcypress/health-data-standards/files/1149161/CMS158_v5_3_Artifacts.zip)
[CMS160_v5_3_Artifacts.zip](https://github.com/projectcypress/health-data-standards/files/1149162/CMS160_v5_3_Artifacts.zip)
[Test134_v5_3_Artifacts.zip](https://github.com/projectcypress/health-data-standards/files/1149163/Test134_v5_3_Artifacts.zip)
[TestCMS177_v5_3_Artifacts.zip](https://github.com/projectcypress/health-data-standards/files/1149164/TestCMS177_v5_3_Artifacts.zip)

JIRA test: https://jira.mitre.org/browse/BONNIE-823


lib/health-data-standards/util/hqmf_template_helper.rb
```diff
@@ -26,6 +27,9 @@ def self.template_id_by_definition_and_status(definition, status, negation=false
         when "r2"
           kv_pair = template_id_map(version).find {|k, v| v['definition'] == definition &&
                                                  v['status'] == status}
+        when "r2cql"
```
pkmitre: Looks like the r2 and r2cql code blocks are the same, in cases like that you can use commas in the when statement, ie

    when "r2", "r2cql"
      kv_pair = template_id_map(version).find {|k, v| v['definition'] == definition &&
                                                      v['status'] == status}

lib/hqmf-parser/cql/data_criteria.rb
```diff
@@ -0,0 +1,59 @@
...
+      @entry = entry
+      @data_criteria_references = data_criteria_references
+      @occurrences_map = occurrences_map
+      basic_setup
+      @variable = HQMF2::DataCriteriaMethods.extract_variable(@local_variable_name, @id)
+      @field_values = HQMF2::DataCriteriaMethods.extract_field_values(@entry, @negation)
+      @description = extract_description
+      obtain_specific_and_source = HQMF2::SpecificOccurrenceAndSource.new(@entry, @id, @local_variable_name,
+                                                                          @data_criteria_references, @occurrences_map)
+      # Pulling these 5 variables out via destructing
```
pkmitre: Nit, but destructing -> destructuring (yeah, the same typo is in the base class :-)

lib/hqmf-parser/cql/data_criteria.rb
```diff
@@ -0,0 +1,59 @@
...
+      @source_data_criteria,
+        @source_data_criteria_root,
+        @source_data_criteria_extension,
+        @specific_occurrence,
+        @specific_occurrence_const = obtain_specific_and_source.extract_specific_occurrences_and_source_data_criteria
+      extract_definition_from_template_or_type
+      post_processing
+    end
+
+    def clone
```
pkmitre: Could you get rid of this entire clone method by changing the first line of the clone method in the base DataCriteria class from

    other = DataCriteria.new(@entry, @data_criteria_references, @occurrences_map)

to

    other = self.class.new(@entry, @data_criteria_references, @occurrences_map)

(It looks like the only different is that first line)

lib/hqmf-parser/cql/data_criteria.rb
```diff
@@ -0,0 +1,59 @@
+module HQMF2CQL
+  # Represents a data criteria specification
+  class DataCriteria < HQMF2::DataCriteria
+    include HQMF2CQL::DataCriteriaPostProcessing, HQMF2CQL::DataCriteriaTypeAndDefinitionExtraction
+
+    # Create a new instance based on the supplied HQMF entry
+    # @param [Nokogiri::XML::Element] entry the parsed HQMF entry
+    def initialize(entry, data_criteria_references = {}, occurrences_map = {})
```
pkmitre: Looks like this initialize code is the same as the base class, why does it need to be redefined?

lib/hqmf-parser/cql/data_criteria_helpers/dc_definition_from_template_or_type_extract.rb
```diff
@@ -0,0 +1,79 @@
...
+        # that" approach is an artifact of the template oids changing as of
+        # MAT 5.3; we want to support measures exported using 5.3, but also
+        # measures that were exported using previous versions of the MAT.
+
+        # Try a lookup using the older template oids.
+        defs = HQMF::DataCriteria.definition_for_template_id(template_id, 'r2')
+        
+        # If the old template oids didn't work, try a lookup using the newer
+        # template oids.
+        defs = HQMF::DataCriteria.definition_for_template_id(template_id, 'r2cql') unless defs
```
pkmitre: Would changing the order to look up using new template oids first and fall back to the old oids prevent potential issues where the oid sets overlap?

lib/health-data-standards/util/hqmf_template_helper.rb
holmesie: Good point, fixed.

lib/hqmf-parser/cql/data_criteria.rb
holmesie: The code here is the same as the base class, but the method calls it makes are to versions that have been modified and included within the HQMF2CQL module. Is there a way to force override versions of methods between different modules without having to modify the code in the original module?

lib/hqmf-parser/cql/data_criteria.rb
holmesie: Aha, missed that - fixed.

lib/hqmf-parser/cql/data_criteria.rb
holmesie: I didn't even consider that, I like it; fixed.

lib/hqmf-parser/cql/data_criteria_helpers/dc_definition_from_template_or_type_extract.rb
holmesie: Luckily they don't where it is important

holmesie: Scratch that, I changed my mind about this; I think that makes sense even if it isn't the case at the moment.

alexanderelliott121 (approved):

lib/hqmf-parser/cql/data_criteria.rb
pkmitre: In ruby, an object only ever has one definition of a method; if a class has one definition of a method through inheritance, and then another definition provided by including a module, the module definition overrides the inherited class definition, and it doesn't matter whether the method calling the overridden method comes from the base class or the child class. Here's something that shows what I mean:

    module Zoom1
      def go 
        puts 'Zoom1'
      end
    end

    module Zoom2
      def go
        puts 'Zoom2'
      end
    end

    class Foo
      include Zoom1
      def initialize
        go
      end
    end

    class FooKid < Foo
      include Zoom2
    end

    Foo.new # Zoom1

    FooKid.new # Zoom2

That said... if you tried it, and it didn't work that way, there's likely something happening that I didn't consider :->

lib/hqmf-parser/2.0/data_criteria.rb
```diff
@@ -148,7 +148,7 @@ def handle_derived_specific_occurrence_variable
     # not give us the correct SDC data when we are trying to recreate since we are looping back through
     # the same data criteria before it has finished processing: See: DocUtilities.extract_source_data_criteria
     def clone
-      other = DataCriteria.new(@entry, @data_criteria_references, @occurrences_map)
+      other = self.class.new(@entry, @data_criteria_references, @occurrences_map)
```
pkmitre: Maybe a comment indicating that this approach is to allow child classes to successfully use clone...

lib/hqmf-parser/cql/data_criteria.rb
holmesie: Okay, that's very clear; All I need is the includes. Just gave it a whirl and things are working as expected. Fixed.

lib/hqmf-parser/2.0/data_criteria.rb
holmesie: Added.

pkmitre (approved):

PR #515, opened 2017-07-15
Cql4bonnie 814

lukoktonos:
JIRA Test: https://jira.mitre.org/browse/BONNIE-838


lukoktonos:
Manual regression test passed.

lib/health-data-standards/models/encounter.rb
```diff
@@ -22,6 +22,8 @@ class Encounter < Entry
   alias :transfer_from= :transferFrom=
   alias :principal_diagnosis :principalDiagnosis
   alias :principal_diagnosis= :principalDiagnosis=
+  alias :admissionSource :admission_source
+  alias :admissionSource :admission_source=
```
pkmitre: The second alias should be

    alias :admissionSource= :admission_source=

instead of

    alias :admissionSource :admission_source=

(there's a missing `=` after the `:admissionSource`)

lib/health-data-standards/models/admission_source.rb
```diff
@@ -0,0 +1,9 @@
+class AdmissionSource
+  include Mongoid::Document
+  include Mongoid::Attributes::Dynamic
+  
+  embedded_in :encounter, class_name: "Encounter", inverse_of: nil
+  
+  field :code, type: String
```
pkmitre, line 7: In cql_qdm_patientapi file app/assets/javascripts/datatypes/encounter.js.coffee line 141 it looks like it's expecting `admissionSource` to have a `code_system` in addition to a `code` (though I don't see anywhere `code_system` would get set... or even `code`).

lukoktonos: in bonnie/lib/measures/patient_builder.rb on line 323 there is `entry.send("#{field_accessor}=", field_value)`.  `field_value` for this case looks like `{"code_system"=>"SNOMED-CT", "code"=>"18095007", "title"=>"Hospital Settings"}`, so I think that is where `code` and `code_system` are coming from.

pkmitre: Do we then want a `code_system` field on `AdmissionSource`?

alexanderelliott121 (approved):

lib/health-data-standards/models/admission_source.rb
lukoktonos: Good call.  I've updated with that field and added `title` as well.  I don't think this bug will need either of those fields, but it's probably best to be complete.

holmesie (approved):
Ran JIRA test, ran regression test - looks good!

edeyoung:
pete has said that he has ok'd the cocde

holmesie (approved):

PR #521, opened 2017-08-02
Removed Radiation Dose and Radiation Duration per QDM 5.3

c-monkey:
See https://github.com/projecttacoma/cql_qdm_patientapi/pull/59 for details


mayerm94 (approved):
These changes look good to me

holmesie (approved):
Changes look good

edeyoung:
To support maintaining backwards compatibility, this change will not be incorporated. Instead, we'll add TODOs for what items should be removed for QDM 5.3. This will be a separate task.

PR #523, opened 2017-08-03
Cql4bonnie qdm5 3 937

lukoktonos:
JIRA Test: https://jira.mitre.org/browse/BONNIE-961

### DIFF
--- a/lib/health-data-standards/models/encounter.rb
+++ b/lib/health-data-standards/models/encounter.rb
@@ -3,9 +3,9 @@ class Encounter < Entry
   field :dischargeDisposition, as: :discharge_disposition, type: Hash
   field :admitTime, as: :admit_time, type: Integer
   field :dischargeTime, as: :discharge_time, type: Integer
+  field :diagnosis, type: Hash
 
   embeds_one :principalDiagnosis, class_name: "EncounterPrincipalDiagnosis"
-  embeds_one :diagnoses, class_name: "EncounterPrincipalDiagnosis"
   
   embeds_one :transferTo, class_name: "Transfer"
   embeds_one :transferFrom, class_name: "Transfer"
--- a/lib/health-data-standards/models/encounter.rb
+++ b/lib/health-data-standards/models/encounter.rb
@@ -5,7 +5,7 @@ class Encounter < Entry
   field :dischargeTime, as: :discharge_time, type: Integer
 
   embeds_one :principalDiagnosis, class_name: "EncounterPrincipalDiagnosis"
-  embeds_one :diagnosis, class_name: "EncounterPrincipalDiagnosis"
+  embeds_one :diagnoses, class_name: "EncounterPrincipalDiagnosis"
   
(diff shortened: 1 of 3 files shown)


lib/hqmf-parser/cql/document_helpers/doc_population_helper.rb
```diff
@@ -73,9 +73,7 @@ def extract_populations_cql_map
               populations_cql_map[criteria_id] = []
             end
             cql_statement = cql_statement.to_s.delete('\\"')
-            unless populations_cql_map[criteria_id].include? cql_statement
```
mayerm94, line 76: What was the motivation for removing this?

edeyoung: hm... it looks like the cql4bonnie-QDM_5-3 branch needs the latest merge from cql4bonnie (this was a bug fix a bit ago), and it looks like luke originally branched from cql4bonnie.

edeyoung: I just did that. the PR doesn't seem to be updating though.

jbradl11 (approved):
This looks good, if we are ok with breaking backwards compatibility. By changing the model, I believe this will break previous versions. Test passed.

mayerm94 (approved):

lib/health-data-standards/models/encounter.rb
```diff
@@ -5,7 +5,7 @@ class Encounter < Entry
   field :dischargeTime, as: :discharge_time, type: Integer
 
   embeds_one :principalDiagnosis, class_name: "EncounterPrincipalDiagnosis"
-  embeds_one :diagnosis, class_name: "EncounterPrincipalDiagnosis"
+  embeds_one :diagnoses, class_name: "EncounterPrincipalDiagnosis"
```
edeyoung: does this not need to change to `embeds_many`?

edeyoung (changes requested):
respond to `embeds_many` question

lib/health-data-standards/models/encounter.rb
lukoktonos: The encounter has one 'diagnoses' field, which is an array of diagnoses, rather than embedding many singular diagnosis fields.

edeyoung: I'm blanking on this. Is it an array simply because we assign an array to it at some point? Have you confirmed that it actually gets stored in the db as an array? I don't see any changes below that indicate that diagnoses is now an array, which is why i'm asking.

thanks!

lukoktonos: I looked to how components was handled in health_data_standards, which is like this:
image
I thought that embeds_one would be similar to field in this case.

The database shows that diagnoses is of the Collection type that Matt made:
image

The values represent the individual diagnosis elements.

edeyoung (approved):

PR #525, opened 2017-08-04
Add logic that will find where there is only a negated data criterion…

mulcahyk:
Moves the logic that checks for only negated data criteria from bonnie-bundler into HDS.
Goes with https://github.com/projecttacoma/bonnie_bundler/pull/101

JIRA test: https://jira.mitre.org/browse/BONNIE-839

### DIFF
--- a/lib/hqmf-parser/cql/data_criteria.rb
+++ b/lib/hqmf-parser/cql/data_criteria.rb
@@ -28,7 +28,30 @@ module HQMF2CQL
       title_match = title.match(/(.*) \w+ [Vv]alue [Ss]et/)
       @title = title_match[1] if title_match && title_match.length > 1
       
-      @description = "#{@description}: #{title}" if (/: #{title}$/).match(@description).nil?
+      @description = "#{@description}: #{title}"
     end
+
+    # In certain situations it is necessary to have a negated data criterion 
+    # copied to a "positive" form.
+    def make_criterion_positive
+      @negation = false
+
+      # Remove negation from description
+      # sometimes "Not Done" used: "Communication: From Provider To Patient, Not Done"
+      # should transform to "Communication: From Provider To Patient"
+      @description.gsub!(', Not Done', '')
+      
+      # sometimes just "Not" used: "Encounter, Not Performed"
+      # should transform to "Encounter, Performed"
+      @description.gsub!(', Not', ',')
+
+      @source_data_criteria = 'Derived from ' + @source_data_criteria
+            
+      # Looking to remove the word 'Not'.  Using lookahead and lookbehind in the regex
+      # criterion.id = criterion.id.gsub(/(?<=[a-z])Not(?=[A-Z])/, '') + '_spoof'
+      @id = @id.gsub(/(?<=[a-z])Not(?=[A-Z])/, '') + '_spoofed'
+
+    end
+    
   end
 end
--- a/lib/hqmf-parser/cql/data_criteria.rb
+++ b/lib/hqmf-parser/cql/data_criteria.rb
@@ -27,8 +27,8 @@ module HQMF2CQL
       # remove * Value Set from title
       title_match = title.match(/(.*) \w+ [Vv]alue [Ss]et/)
       @title = title_match[1] if title_match && title_match.length > 1
-
-      @description = "#{@description}: #{title}"
+      
+      @description = "#{@description}: #{title}" if (/: #{title}$/).match(@description).nil?
     end
   end
 end
--- a/lib/hqmf-parser/cql/document.rb
+++ b/lib/hqmf-parser/cql/document.rb
@@ -42,6 +42,62 @@ module HQMF2CQL
         @data_criteria << dc
         @source_data_criteria << sdc
       end
+      make_positive_entry
+    end
+
+    # This method is needed for situations when there is a only
+    # a negated version of a data criteria.  Bonnie will only
+    # show the affirmative version of data criteria.  This method
+    # will create an affirmative version of a data criteria when there
+    # is only the negative one in the HQMF.
+    def make_positive_entry
+      negated_criteria = {}
+      description_hash = {}
+      data_criteria_index_lookup = {}
+      # Find the criteria that are negated
+      # At the same time build a hash of criteria and their descriptions
+      @data_criteria.each_with_index do |criterion, source_index|
+        negated_criteria[source_index] = criterion if criterion.negation
+        # Using the criterion name as there might be multiples of the same description
+        # description_hash[criterion] = detail['description']
+        data_criteria_index_lookup[source_index] = [criterion.code_list_id, criterion.definition, criterion.status, criterion.negation]
+      end
+
+      unless negated_criteria.count == 0
+        negated_criteria.each do |negated_index, criterion|
+          # Check if there is a criterion has the affirmative description
+          unless data_criteria_index_lookup.value?([criterion.code_list_id, criterion.definition, criterion.status, false])
+            # Make the new name based on the title, definition, and status
+            spoofed_title = criterion.title.gsub(' ', '')
+            spoofed_def = criterion.definition ? criterion.definition.split.map(&:capitalize).join('') : ''
+            spoofed_status = criterion.status ? criterion.status.split.map(&:capitalize).join('') : ''
+            spoofed_criterion_name = spoofed_title + '_' + spoofed_def + spoofed_status + '_spoofed'
+            criterion.instance_variable_set(:@negation, false)
+
+            description = criterion.description
+            # Remove negation from description
+            # sometimes "Not Done" used: "Communication: From Provider To Patient, Not Done"
+            # should transform to "Communication: From Provider To Patient"
+            description = description.gsub(', Not Done', '')
+            
+            # sometimes just "Not" used: "Encounter, Not Performed"
+            # should transform to "Encounter, Performed"
+            description = description.gsub(', Not', ', ')
+            
+            criterion.instance_variable_set(:@description, description)
+            criterion.instance_variable_set(:@source_data_criteria, 'Derived from ' + criterion.source_data_criteria)
+            
+            # Looking to remove the word 'Not'.  Using lookahead and lookbehind in the regex
+            criterion.id = criterion.id.gsub(/(?<=[a-z])Not(?=[A-Z])/, '') + '_spoof'
+            
+            @data_criteria << criterion
+            sdc = criterion.clone
+            sdc.id += '_source'
+            @source_data_criteria << sdc
+          end
+        end
+
+      end
     end
 
   end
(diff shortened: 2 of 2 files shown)


lib/hqmf-parser/cql/document.rb
```diff
@@ -42,6 +42,62 @@ def extract_criteria
...
+    def make_positive_entry
+      negated_criteria = {}
+      description_hash = {}
+      data_criteria_index_lookup = {}
+      # Find the criteria that are negated
+      # At the same time build a hash of criteria and their descriptions
+      @data_criteria.each_with_index do |criterion, source_index|
+        negated_criteria[source_index] = criterion if criterion.negation
+        # Using the criterion name as there might be multiples of the same description
+        # description_hash[criterion] = detail['description']
```
holmesie: Is this part of the comment? Or just commented out code?

lib/hqmf-parser/cql/document.rb
```diff
@@ -42,6 +42,62 @@ def extract_criteria
...
+
+            description = criterion.description
+            # Remove negation from description
+            # sometimes "Not Done" used: "Communication: From Provider To Patient, Not Done"
+            # should transform to "Communication: From Provider To Patient"
+            description = description.gsub(', Not Done', '')
+            
+            # sometimes just "Not" used: "Encounter, Not Performed"
+            # should transform to "Encounter, Performed"
+            description = description.gsub(', Not', ', ')
```
holmesie: With this `gsub`, `Encounter, Not Performed` would be turned into `Encounter,  Performed` (two spaces between the `,` and `Performed`). I think it should be `description = description.gsub(', Not', ',')` instead.

lib/hqmf-parser/cql/data_criteria.rb
```diff
@@ -27,8 +27,8 @@ def retrieve_title_and_description_for_model
       # remove * Value Set from title
       title_match = title.match(/(.*) \w+ [Vv]alue [Ss]et/)
       @title = title_match[1] if title_match && title_match.length > 1
-
-      @description = "#{@description}: #{title}"
+      
+      @description = "#{@description}: #{title}" if (/: #{title}$/).match(@description).nil?
```
holmesie: Why this change? Seems to make sense, but just am curious.

lib/hqmf-parser/cql/document.rb
```diff
@@ -42,6 +42,62 @@ def extract_criteria
...
+
+      unless negated_criteria.count == 0
+        negated_criteria.each do |negated_index, criterion|
+          # Check if there is a criterion has the affirmative description
+          unless data_criteria_index_lookup.value?([criterion.code_list_id, criterion.definition, criterion.status, false])
+            # Make the new name based on the title, definition, and status
+            spoofed_title = criterion.title.gsub(' ', '')
+            spoofed_def = criterion.definition ? criterion.definition.split.map(&:capitalize).join('') : ''
+            spoofed_status = criterion.status ? criterion.status.split.map(&:capitalize).join('') : ''
+            spoofed_criterion_name = spoofed_title + '_' + spoofed_def + spoofed_status + '_spoofed'
```
holmesie: I'm not seeing where this variable is being used.

lib/hqmf-parser/cql/document.rb
holmesie: Also, you can do a `description.gsub!(', Not', ',')` instead here - not a biggie, but is a nice shortcut.

lib/hqmf-parser/cql/data_criteria.rb
mulcahyk: If I don't do this it will add a second : #{title}.  It has to do with the cql version of this file inheriting from the qdm version of the file.  I couldn't find where the first `retrieve_title_and_description_for_model` was being called.

lib/hqmf-parser/cql/document.rb
mulcahyk: Old comment.  Removing.

lib/hqmf-parser/cql/data_criteria.rb
edeyoung: where did you see this? Was this during debugging? Is there a change in the code that now makes this get called multiple times?

I'm very confident that in practice, this only gets called once.

lib/hqmf-parser/cql/document.rb
```diff
@@ -42,6 +42,62 @@ def extract_criteria
...
+      negated_criteria = {}
+      description_hash = {}
+      data_criteria_index_lookup = {}
+      # Find the criteria that are negated
+      # At the same time build a hash of criteria and their descriptions
+      @data_criteria.each_with_index do |criterion, source_index|
+        negated_criteria[source_index] = criterion if criterion.negation
+        # Using the criterion name as there might be multiples of the same description
+        # description_hash[criterion] = detail['description']
+        data_criteria_index_lookup[source_index] = [criterion.code_list_id, criterion.definition, criterion.status, criterion.negation]
```
edeyoung: why is this a hash instead of an array? It never gets used effectively as a hash (using the key for quick access).

lib/hqmf-parser/cql/data_criteria.rb
mulcahyk: I thought it was strange as well.  I actually only saw it on my final round of testing when I was making sure that everything for the spoofed entries matched the nature entries.  As far as I can tell I didn't change anything that would make this get called multiple times.  Perhaps there is something in the onion a few layers down that I just don't know about..?

edeyoung: can you remove this change, try again, and document exactly what you're seeing and what was executed to see this (e.g., through an automated test, through starting the server and uploading a measure in bonnie, through a debugger view, etc.)?

Thank will give me greater comfort for this change.

c-monkey (approved):
Passed https://jira.mitre.org/browse/BONNIE-839. Code is basically the same as source.

c-monkey:
Can we create an automated test for this? Should we?

lib/hqmf-parser/cql/document.rb
c-monkey: actually all 4 of the variables with the prefix "spoofed" are obsolete, please remove.

lib/hqmf-parser/cql/document.rb
```diff
@@ -42,6 +42,60 @@ def extract_criteria
...
+      data_criteria_index_lookup = {}
+      # Find the criteria that are negated
+      # At the same time build a hash of all criteria and their code_list_id, definition, status, and negation status
+      @data_criteria.each_with_index do |criterion, source_index|
+        negated_criteria[source_index] = criterion if criterion.negation
+        data_criteria_index_lookup[source_index] = [criterion.code_list_id, criterion.definition, criterion.status, criterion.negation]
+      end
+
+      unless negated_criteria.count == 0
+        negated_criteria.each do |negated_index, criterion|
```
c-monkey: `negated_index` is not used, how about change `negated_criteria` to an array?

lib/hqmf-parser/cql/document.rb
```diff
@@ -42,6 +42,60 @@ def extract_criteria
...
+          # Check if there is a criterion with the same OID, definition and status BUT that isn't negated
+          unless data_criteria_index_lookup.value?([criterion.code_list_id, criterion.definition, criterion.status, false])
+            # Make the new name based on the title, definition, and status
+            spoofed_title = criterion.title.gsub(' ', '')
+            spoofed_def = criterion.definition ? criterion.definition.split.map(&:capitalize).join('') : ''
+            spoofed_status = criterion.status ? criterion.status.split.map(&:capitalize).join('') : ''
+            spoofed_criterion_name = spoofed_title + '_' + spoofed_def + spoofed_status + '_spoofed'
+            criterion.instance_variable_set(:@negation, false)
```
c-monkey: This changes the original criterion in @data_criteria. Did you mean to instead make a copy? If not, then the code below `@data_criteria << criterion` doesn't do anything

lib/hqmf-parser/cql/document.rb
```diff
@@ -42,6 +42,60 @@ def extract_criteria
...
+
+            description = criterion.description
```
c-monkey: `description` is a pointer to the field, so changing it updates the value in criterion.description

c-monkey (changes requested):
Sorry I approved this before without a proper code inspection.

lib/hqmf-parser/cql/document.rb
```diff
@@ -51,45 +51,24 @@ def extract_criteria
...
-            
-            criterion.instance_variable_set(:@description, description)
-            criterion.instance_variable_set(:@source_data_criteria, 'Derived from ' + criterion.source_data_criteria)
-            
-            # Looking to remove the word 'Not'.  Using lookahead and lookbehind in the regex
-            criterion.id = criterion.id.gsub(/(?<=[a-z])Not(?=[A-Z])/, '') + '_spoof'
-            
-            @data_criteria << criterion
-            sdc = criterion.clone
+            smiley_face = criterion.clone
```
edeyoung: you still have `smiley_face` :)

lib/hqmf-parser/cql/document.rb
```diff
@@ -42,6 +42,39 @@ def extract_criteria
...
+
+    # This method is needed for situations when there is a only
+    # a negated version of a data criteria.  Bonnie will only
+    # show the affirmative version of data criteria.  This method
+    # will create an affirmative version of a data criteria when there
+    # is only the negative one in the HQMF.
+    def make_positive_entry
+      negated_criteria = []
+      description_hash = {}
+      data_criteria_index_lookup = {}
```
edeyoung: looks like my question about this being a hash instead of an array got lost.

this is treated like an array (only looking things up by value). Why not make it be an array? It's not used effectively as a hash (referenced by key).

lib/hqmf-parser/cql/document.rb
```diff
@@ -42,6 +42,39 @@ def extract_criteria
...
+      # Find the criteria that are negated
+      # At the same time build a hash of all criteria and their code_list_id, definition, status, and negation status
+      @data_criteria.each_with_index do |criterion, source_index|
+        negated_criteria << criterion if criterion.negation
+        data_criteria_index_lookup[source_index] = [criterion.code_list_id, criterion.definition, criterion.status, criterion.negation]
+      end
+
+      unless negated_criteria.count == 0
```
edeyoung: you don't need this check. an empty array just doesn't have element to iterate over.

lib/hqmf-parser/cql/document.rb
```diff
@@ -42,6 +42,39 @@ def extract_criteria
...
+    end
+
+    # This method is needed for situations when there is a only
+    # a negated version of a data criteria.  Bonnie will only
+    # show the affirmative version of data criteria.  This method
+    # will create an affirmative version of a data criteria when there
+    # is only the negative one in the HQMF.
+    def make_positive_entry
+      negated_criteria = []
+      description_hash = {}
```
edeyoung: it doesn't look like this gets used.

c-monkey (approved):
Test passes: https://jira.mitre.org/browse/BONNIE-839

My comments were addressed.

holmesie (approved):
Test passes, looks good!

PR #530, opened 2017-08-16
Update HDS models and tests to reflect new Facility Collection

lukoktonos:
JIRA Test: https://jira.mitre.org/browse/BONNIE-968

Related PRs:
- https://github.com/projecttacoma/cql_qdm_patientapi/pull/72
- https://github.com/projecttacoma/bonnie/pull/819
- https://github.com/projectcypress/health-data-standards/pull/530

### DIFF
--- a/lib/health-data-standards/models/adverse_event.rb
+++ b/lib/health-data-standards/models/adverse_event.rb
@@ -3,6 +3,4 @@ class AdverseEvent < Entry
   field :type, type: Hash
   field :severity, type: Hash
   field :facility, type: Hash
-  # TODO: encounter includes a `facility` object. should that be done here? all we care about is the location.
-  #field :facilityLocation, as: :facility_location, type: String
 end
--- a/lib/health-data-standards/models/adverse_event.rb
+++ b/lib/health-data-standards/models/adverse_event.rb
@@ -2,7 +2,7 @@
 class AdverseEvent < Entry
   field :type, type: Hash
   field :severity, type: Hash
+  field :facility, type: Hash
   # TODO: encounter includes a `facility` object. should that be done here? all we care about is the location.
   #field :facilityLocation, as: :facility_location, type: String
-  embeds_one :facility
 end
--- a/lib/health-data-standards/models/encounter.rb
+++ b/lib/health-data-standards/models/encounter.rb
@@ -4,13 +4,13 @@ class Encounter < Entry
   field :admitTime, as: :admit_time, type: Integer
   field :dischargeTime, as: :discharge_time, type: Integer
   field :diagnosis, type: Hash # This is a Hash to align with the 'Collection' object
+  field :facility, type: Hash
 
   embeds_one :principalDiagnosis, class_name: "EncounterPrincipalDiagnosis"
   
   embeds_one :transferTo, class_name: "Transfer"
   embeds_one :transferFrom, class_name: "Transfer"
 
-  embeds_one :facility
   embeds_one :admission_source, class_name: "AdmissionSource"
   embeds_one :reason, class_name: "Entry"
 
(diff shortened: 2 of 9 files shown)


lib/health-data-standards/models/adverse_event.rb
```diff
@@ -2,7 +2,7 @@
 class AdverseEvent < Entry
   field :type, type: Hash
   field :severity, type: Hash
+  field :facility, type: Hash
   # TODO: encounter includes a `facility` object. should that be done here? all we care about is the location.
```
c-monkey: this comment is obsolete, should be removed

lib/health-data-standards/models/encounter.rb
```diff
@@ -28,11 +28,24 @@ class Encounter < Entry
...
     
     self.admitTime = (self.admitTime.nil?) ? nil : self.admitTime + date_diff
     self.dischargeTime = (self.dischargeTime.nil?) ? nil : self.dischargeTime + date_diff
   end
 
+  def shift_facility_dates(date_diff)
+    # Facility is now stored as a Hash; shift its dates by updating its start_time and end_time values.
+    if self.facility
+      if self.facility['start_time']
+        self.facility['start_time'] = self.facility['start_time'] + date_diff
```
c-monkey: I think this would be cleaner if you used an assignment operator, i.e. `+=`. See http://www.techotopia.com/index.php/Ruby_Operators#Ruby_Assignment_Operators for description.

Also, this codebase frequently uses trailing expression modifiers so you might rearrange the code that way. I don't feel strongly about this and across the codebase we're inconsistent. See http://ruby-doc.org/core-2.1.2/doc/syntax/control_expressions_rdoc.html#label-Modifier+if+and+unless

So this block might turn into:
```
if self.facility
  self.facility['start_time'] += date_diff if self.facility['start_time']
  self.facility['end_time'] += date_diff if self.facility['end_time']
```

lib/health-data-standards/models/encounter.rb
```diff
@@ -28,11 +28,24 @@ class Encounter < Entry
...
+      end
+
+      if self.facility['end_time']
+        self.facility['end_time'] = self.facility['end_time'] + date_diff
```
c-monkey: ditto

lib/health-data-standards/models/encounter.rb
```diff
@@ -28,11 +28,24 @@ class Encounter < Entry
   def shift_dates(date_diff)
     super
     if self.facility
```
c-monkey, line 30: this check is now redundant and can be removed - already performed in shift_facility_dates function

mayerm94 (approved):
Tested, jira test passes

test/unit/import/c32/encounter_importer_test.rb
```diff
@@ -1,4 +1,3 @@
-
 require 'test_helper'
 
 class EncounterImporterTest < Minitest::Test
```
c-monkey, line 3: you didn't add an assertion for the new field  in facility`type`

test/unit/models/date_shift_test.rb
```diff
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'byebug'
```
c-monkey: please remove this debugger statement

test/unit/models/date_shift_test.rb
```diff
@@ -56,9 +57,9 @@ def test_encounter_shift
 				facility_values.each do |fac_vals|
 					e_vals = vals.merge enc_vals
 			  	entry = Encounter.new(e_vals)
-			  	entry.facility = Facility.new(fac_vals)
+				  entry.facility = Facility.new(fac_vals).as_json()
```
c-monkey, line 59: Oh, this file has TABs, no wonder the whitespace is screwy! I hate it. Don't fix that in this PR, though! We should make a separate PR to fix this outlier file (I did a global search, this is the only one). Not a priority for this sprint.

c-monkey (changes requested):

test/unit/import/c32/encounter_importer_test.rb
lukoktonos: I realized `type` was unused when trying to write this test, so I removed it.

lib/health-data-standards/models/encounter.rb
lukoktonos: I misread this and removed it from the shift_facility_dates function instead, but in any case there is only 1 check now.

test/unit/models/date_shift_test.rb
lukoktonos: Ah, that explains why the indentation seemed so weird.

lib/health-data-standards/models/encounter.rb
c-monkey: your tweak to remove the empty check here means theoretically one could call shift_facility_dates() with an empty self.facility and cause an error. Why not just move the 3 lines of that new function here and remove it?

c-monkey (approved):
changes addressed my comments

manual JIRA test 968 passed

PR #533, opened 2017-08-31
Cql4bonnie 1028

mayerm94:
Test with Bonnie cql4bonnie_1028

### DIFF
--- a/lib/hqmf-model/types.rb
+++ b/lib/hqmf-model/types.rb
@@ -487,4 +487,47 @@ module HQMF
     end  
   end
   
+  class Ratio
+    include HQMF::Conversion::Utilities
+    attr_accessor :type, :numerator, :denominator
+    
+    # @param [String] type
+    # @param [Value] numerator
+    # @param [Value] denominator
+  
+    def initialize(type, numerator, denominator)
+      @type = type || 'RT'
+      @numerator = numerator
+      @denominator = denominator
+    end
+    
+    def self.from_json(json)
+      type = json["type"] if json["type"]
+      numerator = json["numerator"] if json["numerator"]
+      denominator = json["denominator"] if json["denominator"]
+      HQMF::Ratio.new(type, numerator, denominator)
+    end
+  
+    def numerator?
+      @numerator
+    end
+
+    def denominator?
+      @denominator
+    end
+
+    def to_json
+      build_hash(self, [:type, :numerator, :denominator])
+    end
+    
+    def stringify
+      "#{numerator? '\''}#{denominator}"
+    end
+
+    def ==(other)
+      check_equality(self, other)
+    end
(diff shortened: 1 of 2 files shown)


lib/hqmf-model/types.rb
```diff
@@ -487,4 +487,47 @@ def to_json
...
+    def denominator?
+      @denominator
+    end
+
+    def to_json
+      build_hash(self, [:type, :numerator, :denominator])
+    end
+    
+    def stringify
+      "#{numerator? '\''}#{denominator}"
```
lukoktonos: What is the stringified ratio supposed to look like?  I believe this will turn numerator=5 and denominator=6 into "5 '6" ? Unless I'm misinterpreting the ruby syntax

lib/hqmf-model/types.rb
```diff
@@ -487,4 +487,47 @@ def to_json
...
   end
   
+  class Ratio
+    include HQMF::Conversion::Utilities
+    attr_accessor :type, :numerator, :denominator
+    
+    # @param [String] type
+    # @param [Value] numerator
+    # @param [Value] denominator
+
```
holmesie: Remove this empty line.

lib/hqmf-model/types.rb
```diff
@@ -487,4 +487,47 @@ def to_json
...
+    end
+    
+    def self.from_json(json)
+      type = json["type"] if json["type"]
+      numerator = json["numerator"] if json["numerator"]
+      denominator = json["denominator"] if json["denominator"]
+      HQMF::Ratio.new(type, numerator, denominator)
+    end
+  
+    def numerator?
```
holmesie, line 510: This is a bit confusing to me. Is the purpose of this method to say if this `Ratio` has a numerator? If so, since the `?` means this is a predicate method, I would recommend making sure you are actually returning a boolean, rather than just relying on the truthiness of `@numerator`. An easy way of doing that is to prepend `@numerator` with `!!`.

If the purpose of this was to actually return the numerator value, since you delcared this variable with `attr_accessor`, you get the read/write methods for free.

Edit: I just noticed it is done this way all over this file. With that in mind, I think just being consistent is fine.

lib/hqmf-model/types.rb
```diff
@@ -487,4 +487,47 @@ def to_json
...
+      @numerator
+    end
+
+    def denominator?
```
holmesie, line 514: See comment about `def numerator?`.

holmesie (changes requested):

holmesie (approved):

edeyoung:
RATIO WORK NO LONGER BEING USED AS OF 9/1 DISCUSSION IN HQMF CQL TOOLING CALL

PR #543, opened 2017-09-28
BONNIE-1072 VSAC Call Changes

alexanderelliott121:
Intent of PR is to bring the HDS vsac API into conformity with the vsac web api.

JIRA: https://jira.mitre.org/browse/BONNIE-1072
Tests: Automated

- [X] This pull request describes *why* these changes were made.
- [X] Tests are included and test edge cases
- [X] Link to JIRA test(s), if used: NA
- [X] JIRA test(s) have been added to sprint, if used NA
- [X] JIRA ticket for this PR: 
- [X] JIRA ticket links to this PR.
- [x] Automated regression test(s) pass (if applicable): 

Reviewer 1: @hossenlopp
- [x] The code is maintainable
- [x] The code reuses code or resources within the project ecosystem where applicable
- [x] The code achieves its purpose
- [x] The tests appropriately test edge cases
- [x] The tests pass: ONLY PASS LOCALLY. Travis errors are possibly a code issue from outside this PR. 

Reviewer 2: @c-monkey
- [x] The code is maintainable
- [x] The code reuses code or resources within the project ecosystem where applicable
- [x] The code achieves its purpose
- [x] The tests appropriately test edge cases
- [x] The tests pass

### DIFF
--- a/lib/health-data-standards/util/vs_api.rb
+++ b/lib/health-data-standards/util/vs_api.rb
@@ -59,8 +59,6 @@ module HealthDataStandards
       # The VSAC V2 API needs a profile to be specified when using includeDraft. Future work on this 
       # class could include a function to fetch the list of profiles from the https://vsac.nlm.nih.gov/vsac/profiles
       # call.
-      DEFAULT_PROFILE = "Most Recent CS Versions"
-      
       def initialize(ticket_url, api_url, username, password, ticket_granting_ticket = nil)
         super(ticket_url, api_url, username, password, ticket_granting_ticket)
       end
@@ -71,7 +69,7 @@ module HealthDataStandards
         profile = options.fetch(:profile, nil)
         effective_date = options.fetch(:effective_date, nil)
         program_name = options.fetch(:program, nil)
-        if profile.nil? && include_draft
+        if (profile.nil? || profile.empty?) && include_draft
           raise MalformedVSQueryError, "Include Draft Specified, no Profile provided"
         end
         params = { id: oid, ticket: get_ticket }
--- a/lib/health-data-standards/util/vs_api.rb
+++ b/lib/health-data-standards/util/vs_api.rb
@@ -4,6 +4,8 @@ module HealthDataStandards
   module Util
     class VSNotFoundError < StandardError
     end
+    class MalformedVSQueryError < StandardError
+    end
 
     class VSApi
       attr_accessor :api_url, :ticket_url, :username, :password
@@ -69,6 +71,9 @@ module HealthDataStandards
         profile = options.fetch(:profile, nil)
         effective_date = options.fetch(:effective_date, nil)
         program_name = options.fetch(:program, nil)
+        if profile.nil? && include_draft
+          raise MalformedVSQueryError, "Include Draft Specified, no Profile provided"
+        end
         params = { id: oid, ticket: get_ticket }
         params[:version] = version if version
         params[:includeDraft] = 'yes' if profile && include_draft
--- a/lib/health-data-standards/util/vs_api.rb
+++ b/lib/health-data-standards/util/vs_api.rb
@@ -55,7 +55,6 @@ module HealthDataStandards
 
     class VSApiV2 < VSApi
       
-      # This default profile is used when the include_draft option is true without a profile specified.
       # The VSAC V2 API needs a profile to be specified when using includeDraft. Future work on this 
       # class could include a function to fetch the list of profiles from the https://vsac.nlm.nih.gov/vsac/profiles
       # call.
--- a/test/unit/util/vs_api_test.rb
+++ b/test/unit/util/vs_api_test.rb
@@ -31,19 +31,15 @@ class VSApiTest < Minitest::Test
       unless valueset_xml_version[valueset_xml[0]] == ""
         stub_request(:get,'https://localhost/vsservice').with(:query =>{:id=>"oid", :ticket=>"ticket", :effectiveDate=>valueset_xml_version[valueset_xml[0]]}).to_return(:body=>valueset_xml_by_date[valueset_xml[0]])
         assert_equal valueset_xml_by_date[valueset_xml[0]], api.get_valueset("oid", valueset_xml_version[valueset_xml[0]])
-        api.get_valueset("oid", valueset_xml_version[valueset_xml[0]]) do |oid,vs|
-          assert_equal "oid", oid
-          assert_equal valueset_xml_by_date[valueset_xml[0]], vs
-        end
+        vs_call_1 = api.get_valueset("oid", valueset_xml_version[valueset_xml[0]])
+        assert_equal valueset_xml_by_date[valueset_xml[0]], vs_call_1
       end
 
       if valueset_xml_version[valueset_xml[0]] == ""
         stub_request(:get,'https://localhost/vsservice').with(:query =>{:id=>"oid", :ticket=>"ticket"}).to_return(:body=>valueset_xml_by_date[valueset_xml[0]])
         assert_equal valueset_xml_by_date[valueset_xml[0]], api.get_valueset("oid")
-        api.get_valueset("oid") do |oid,vs|
-          assert_equal "oid", oid
-          assert_equal valueset_xml_by_date[valueset_xml[0]], vs
-        end
+        vs_call_2 = api.get_valueset("oid")
+        assert_equal valueset_xml_by_date[valueset_xml[0]], vs_call_2
       end
     end
 	end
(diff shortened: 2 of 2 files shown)


lib/health-data-standards/util/vs_api.rb
```diff
@@ -69,6 +71,9 @@ def get_valueset(oid, options = {}, &block)
         profile = options.fetch(:profile, nil)
         effective_date = options.fetch(:effective_date, nil)
         program_name = options.fetch(:program, nil)
+        if profile.nil? && include_draft
```
hossenlopp: Perhaps this should also check if it is an empty string.

test/unit/util/vs_api_test.rb
```diff
@@ -69,6 +69,14 @@ def test_api_v2_with_include_draft_default_profile
     end
   end
   
+  def test_api_v2_with_include_draft_no_profile
+    valueset_xml = %{<?xml version="1.0" encoding="UTF-8"?><RetrieveValueSetResponse xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" cacheExpirationHint="2012-09-20T00:00:00-04:00" xmlns:nlm="urn:ihe:iti:svs:2008" xmlns="urn:ihe:iti:svs:2008"><ValueSet id="2.16.840.1.113883.11.20.9.23" version="Draft"></ValueSet></RetrieveValueSetResponse>}
```
hossenlopp: This line isn't needed for this test.

test/unit/util/vs_api_test.rb
```diff
@@ -69,6 +69,14 @@ def test_api_v2_with_include_draft_default_profile
...
+    api = HealthDataStandards::Util::VSApiV2.new("https://localhost/token", "https://localhost/vsservice", "myusername", "mypassword")
+    assert_raises HealthDataStandards::Util::MalformedVSQueryError do
+      api.get_valueset("oid", include_draft: true)
+    end
+  end
+  
   def test_api_v2_with_include_draft_specified_profile
```
hossenlopp: This existing test does execute the case where both `includeDraft` is `true` and `profile` is provided. It should, if possible, assert that the MalformedVSQueryError isn't thrown.

I noticed an existing issue with this test. All of the asserts for this test are in the block provided to get_valueset. If get_valueset returns early due to an introduced defect, then it won't yield anything and no asserts will occur. Therefore, this test may fail detect defects. The return result of the `get_valueset` function should be used to check the result of the test call.

hossenlopp (changes requested):
I like the addition of the MalformedVSQueryError. It makes sure a hidden default doesn't accidentally get used. There's a few small checks and tests that could be improved.

hossenlopp:
Further suggestion I forgot earlier: The `DEFAULT_PROFILE` should be removed since it's now unused.

lib/health-data-standards/util/vs_api.rb
alexanderelliott121: I think that the options.fetch(:profile, nil) returns a nil if there is an empty string.
I'll do some testing.

c-monkey:
I agree with all of @hossenlopp comments. Definitely put in some negative tests. Also the JIRA ticket says "It doesn't make sense for HDS to have a default profile...It's Bonnie's business to have a default profile." So might as well clean up the function description along with the DEFAULT_PROFILE.

Thanks!

c-monkey (changes requested):
I agree with all of @hossenlopp comments. Definitely put in some negative tests. Also the JIRA ticket says "It doesn't make sense for HDS to have a default profile...It's Bonnie's business to have a default profile." So might as well clean up the function description along with the DEFAULT_PROFILE.

Thanks!

test/unit/util/vs_api_test.rb
alexanderelliott121: I'm going to create a separate JIRA for this, since it would change the scope of the changes, but I agree that is should be addressed.

lib/health-data-standards/util/vs_api.rb
```diff
@@ -59,8 +59,6 @@ class VSApiV2 < VSApi
       # The VSAC V2 API needs a profile to be specified when using includeDraft. Future work on this
```
hossenlopp, line 58: At least the first line of the comment should be removed since you removed the `DEFAULT_PROFILE`.

alexanderelliott121:
@c-monkey I'm not sure I follow what you mean by negative test cases.
Currently I test to make sure that the correct exception is thrown if the wrong arguments are provided.  Is there another negative case I'm missing?

hossenlopp (approved):

test/unit/util/vs_api_test.rb
c-monkey: Please add the link to the separate JIRA.

c-monkey (changes requested):
We decided during the sprint review that if it is expedient, you should just finish up the testing before this PR is accepted. I propose only a few changes:

1) Following up on the request from @hossenlopp I did a little researching, and there is no such thing as asserting that an error wasn't thrown. True, we could create a try/catch block, BUT, if the error were thrown in another test, it would trigger an error, so it's not necessary to test. If desired, we could add a comment to the affect of: `# does not raise HealthDataStandards::Util::MalformedVSQueryError` or somesuch into test test_api_v2_with_include_draft_default_profile.

2) These tests miss a little of the logic, and could be revised to capture all logic permutations, but `get_valueset` is a very simple function. The tests are using the implicit matching of the parameters to the the call to `get_valueset` and the expected parameters in the `.with` of the `stub_request`. Since I was new to the WebMock library, didn't know there was no way to inspect parameters that were sent to the mocked URLs, so this use of WebMock is pretty clever.

3) For completeness, one could test that the stub *isn't* called, i.e.:
```
  def test_api_v2_with_include_draft_no_profile
    the_stub = stub_request(:get,'https://localhost/vsservice')
    api = HealthDataStandards::Util::VSApiV2.new("https://localhost/token", "https://localhost/vsservice", "myusername", "mypassword")
    assert_raises HealthDataStandards::Util::MalformedVSQueryError do
      api.get_valueset("oid", include_draft: true)
    end
    assert_not_requested the_stub
  end
```
Do you have an objection to this? I think it adds clarity.

Thanks,
--Chris

p.s. all tests in HDS and in Bonnie pass.

edeyoung:
What's the status on the regression tests?

c-monkey:
@alexanderelliott121 The travis tests are failing identically to previous PRs. In a previous PR @edeyoung pointed out that it was failing on some people's machines (but not yours or hers, and now this one also doesn't fail on my machine).

Therefore I'm okay with integrating this change.

c-monkey (approved):

c-monkey:
Travis failure ticket is : https://jira.mitre.org/browse/BONNIE-1136

test/unit/util/vs_api_test.rb
```diff
@@ -63,6 +63,7 @@ def test_api_v2_with_include_draft_default_profile
     valueset_xml = %{<?xml version="1.0" encoding="UTF-8"?><RetrieveValueSetResponse xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" cacheExpirationHint="2012-09-20T00:00:00-04:00" xmlns:nlm="urn:ihe:iti:svs:2008" xmlns="urn:ihe:iti:svs:2008"><ValueSet id="2.16.840.1.113883.11.20.9.23" version="Draft"></ValueSet></RetrieveValueSetResponse>}
     stub_request(:get,'https://localhost/vsservice').with(:query =>{:id=>"oid", :ticket=>"ticket", :includeDraft=>"yes", :profile=>"Most Recent CS Versions"}).to_return(:body=>valueset_xml)
     api = HealthDataStandards::Util::VSApiV2.new("https://localhost/token", "https://localhost/vsservice", "myusername", "mypassword")
+    # does not raise HealthDataStandards::Util::MalformedVSQueryError
     api.get_valueset("oid", include_draft: true, :profile=>"Most Recent CS Versions") do |oid,vs|
```
hossenlopp: The result of calling get_valuset should be checked instead of asserting stuff in the block. If get_valueset doesn't actually attempt to get the valueset and just returns nothing, this test won't catch that since the block won't get called at all.

hossenlopp (changes requested):

test/unit/util/vs_api_test.rb
c-monkey: This is a problem in all the tests, then. I know you (@hossenlopp) caught this before but when debugging how the strange stub_request code works, I forgot about this very base case. Since the first parameter `oid` is not changed, we could make a simple test for testing the yield and then in all the rest of these tests just use the normal return value. Removing the, now basically worthless `assert_equal "oid", oid` assertions would be nice for cleaning up the code, too.

So, just change all the calls to `get_valueset` to `vs = ...` and remove the block. Then the one assertion in the block out-dented.

alexanderelliott121: Basically, I think making that change should constitute its own ticket, given its scope.

c-monkey: @hossenlopp 's suggested refactor is trivial, changing 4 tests, each only 2 lines, and identical changes. I disagree that this should be taken out of this PR. The new Bonnie Way is to fix broken tests before we commit.

test/unit/util/vs_api_test.rb
alexanderelliott121: I needed to put it back in for the assert_not_requested.

alexanderelliott121:
Tests have been updated.

c-monkey (approved):
Tested bonnie as well

hossenlopp (approved):

PR #544, opened 2017-09-28
Transfer dates need to be shifted when an encounter is shifted

dczulada:


c-monkey:
@dczulada Please persuade me not to outright fail this with no tests. To make me most happy, there would be automated regression tests that would fail if this change were not implemented.

Thanks,
--Chris

dczulada:
@c-monkey unit test added

c-monkey (approved):
These unit tests are good. Approved.

okeefm (approved):
Tested, transfers are properly being shifted into the reporting period. Approved.

PR #546, opened 2017-10-05
Add support for multiple versions of the same bundle when one is deprecated

rbclark:
Currently if a user tries to add a bundle that has the same version as an existing bundle then the importer will throw an error. In Cypress we have implemented a concept of deprecation which means that a bundle has been mostly deleted and only has the minimal things required for existing tests to be re-run. In order to allow for Cypress users to import a newer version of a bundle we would need to relax this requirement to only check if the conflicting version of the bundle is in a non-deprecated bundle.

### DIFF
--- a/lib/health-data-standards/import/bundle/importer.rb
+++ b/lib/health-data-standards/import/bundle/importer.rb
@@ -31,7 +31,7 @@ module HealthDataStandards
 
             bundle = unpack_bundle(zip_file)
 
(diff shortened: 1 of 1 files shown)


okeefm:
@rbclark @dczulada do we still want this PR open?

rbclark:
I believe @dczulada has said that we are not interested in this being an option.

PR #550, opened 2017-10-13
Dump whole database as a whole because collections collection is empt…

mayerm94:
Dump whole database as a whole because collections collection is empty due to a mongoid update

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -42,9 +42,7 @@ class Minitest::Test
 
   # Delete all collections from the database.
   def dump_database
-    Mongoid.session(:default).collections.each do |collection|
-      collection.drop unless collection.name.include?('system.')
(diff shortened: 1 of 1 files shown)


mayerm94:
JIRA ticket https://jira.mitre.org/browse/BONNIE-1136 

This PR was created to address the fact that automated tests are failing because they were no longer atomic due to the DB not being wiped between tests. This PR makes the DB clear itself correctly.

c-monkey (approved):
These changes are identical to bonnie_bundler and bonnie project changes. Lesson to learn: if changing a support file for unit tests, it might contain copy-and-pasted code from other projects.

okeefm (approved):
Not sure why those were excluded from the drop originally, but the tests pass locally with this change.

PR #556, opened 2017-10-19
updated PR template to include expectations of code diff and code cov…

edeyoung:
…erage expectations and instructions.


dczulada (approved):
Looks good

